PR #49, opened 2015-02-25
Bnc 915571 2

wfeldt:


jreidinger:
LGTM, just is there related bug report for not relaying on nfs server error code?

markkp:
I don't believe there is a bug report for that. It's a problem we ran into with SLES12 where the return code from NFS didn't match anything in particular. So, we decided that if there was an error, there was an error and we wouldn't worry about the exact value of the error code.

wfeldt:
Mark is correct. That line is just a sle12 backport; the bug is also 915571 as qa was testing with nfs.

PR #52, opened 2015-03-06
don't set a default keymap too eagerly

wfeldt:


mvidner:
Code LGTM.
But shouldn't https://bugzilla.suse.com/show_bug.cgi?id=857194 be mentioned somewhere for the changelog? (I don't know if you can generate it from the branch name)

wfeldt:
reworked into pull request #54

PR #53, opened 2015-03-06
don't set a default keymap too eagerly

wfeldt:


wfeldt:
obsoleted by pull request #55

PR #55, opened 2015-03-09
don't set a default keymap too eagerly (bnc #857194)

wfeldt:


wfeldt:
analog to pull request #54

PR #56, opened 2015-03-18
Bnc 901598 2

wfeldt:
As a result of bnc #901598 we would now get a warning dialog when using a plain rpm in the 'dud' option. This is fixed in two steps: (1) count rpms a driver updates internally and (2) verify the rpm signature (duds must be signed unless you run in 'insecure' mode).

### DIFF
--- a/url.c
+++ b/url.c
@@ -1,4 +1,4 @@
-#define _GNU_SOURCE	/* strnlen, getline */
+#define _GNU_SOURCE	/* strnlen, getline, strcasestr */
 
 /*
 
@@ -70,6 +70,10 @@ static void digest_init(url_data_t *url_data);
 static void digest_process(url_data_t *url_data, void *buffer, size_t len);
 static void digest_finish(url_data_t *url_data);
 static int digest_verify(url_data_t *url_data, char *file_name);
+static int warn_signature_failed(char *file_name);
+static int is_gpg_signed(char *file);
+static int is_rpm_signed(char *file);
+static int is_signed(char *file, int check);
 
 
 void url_read(url_data_t *url_data)
@@ -1507,7 +1511,6 @@ int warn_signature_failed(char *file_name)
 /*
   Test if 'file' is a gpg signed file.
   If so, unpack it (replacing 'file') and verify signature.
-  if 'check' is set, update config.sig_failed and show warning to user.
 
   Return values:
     -1: file or gpg not found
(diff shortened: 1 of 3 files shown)


url.c
```diff
@@ -70,6 +70,10 @@ static void digest_init(url_data_t *url_data);
 static void digest_process(url_data_t *url_data, void *buffer, size_t len);
 static void digest_finish(url_data_t *url_data);
 static int digest_verify(url_data_t *url_data, char *file_name);
+static int warn_signature_failed(char *file_name);
+static int is_gpg_signed(char *file);
+static int is_rpm_signed(char *file);
+static int is_signed(char *file, int check);
```
jreidinger: I am curious why file is not const. Do you modify string?

wfeldt: You could make it const; it's rarely done in linuxrc code.

url.c
```diff
@@ -1567,12 +1570,104 @@ int is_gpg_signed(char *file, int check)
...
+      if(strstr(s, " pgp ") || strstr(s, " gpg ")) sig_ok = 1;
+    }
+    err = pclose(f) ? 1 : 0;
+    if(config.debug >= 2) fprintf(stderr, "rpmkeys returned %s\n", err ? "an error" : "ok");
+  }
+
+  str_copy(&cmd, NULL);
+  free(buf);
+
+  if(err != -1) {
```
jreidinger: when it can happen? for me only if popen failed or getline is 0. In such case buf can be NULL and I think it is not needed be free. so it can be shortly exited above with `if (err == -1) return err;`

wfeldt: I don't understand. A short exit is never possible. I want the status logged at the end.

jreidinger: It is not logged if file do not exists and also in case that that type is not rpm.
If you need such style, maybe something like kernel jumps can be useful - https://www.kernel.org/doc/Documentation/CodingStyle Chapter 7: Centralized exiting of functions

wfeldt: It is logged.

jreidinger: fstype util logging it? https://github.com/openSUSE/linuxrc/pull/56/files#diff-e08b038e4a2f364e3c2303cfff9184f8R1603

wfeldt: Ok, if it's not an rpm, the fact is not logged. But I don't see how that matters. I still want the results of the signature checking logged. And please keep coding style out.

jreidinger: ok, it is consistent with rest of code. so LGTM

PR #58, opened 2015-04-20
Fix BOOTIF on PowerNV

k0da:
On PowerNV IPAPPEND looks like just a mac, so it is _NOT_ prefixed
with 01- or whatsoever.

Let's handle that case in linuxrc

Signed-off-by: Dinar Valeev <dvaleev@suse.com>


k0da:
updated

PR #61, opened 2015-04-24
replace internal nfs mount code with call to mount.nfs (bnc #926982)

wfeldt:
That's basicalyy the sle12 way, code backported from there.


jreidinger:
looks really better then before. LGTM

PR #62, opened 2015-05-06
Code documentation improvements

mvidner:
Related to code I read while trying to understand and fix
https://bugzilla.suse.com/show_bug.cgi?id=925173

I think the factual and typo fixes are fairly uncontroversial, 

Doxygen may not be the best tool/markup, I am open to suggestions how to do it better.

Also I didn't like the mix of APIs using integers as booleans (1==success) and error codes (0==success). How about we start using `bool` for the former and  something else* for the latter? (* making up `error_t`, I can't find a [name for this convention](http://stackoverflow.com/questions/1882047/convention-result-and-code-error-c-int-foo))


wfeldt:
This would be ok for master; I don't see any real value for it in sle11.

mvidner:
I have rebased this from `sl_11.1` onto `master` but AFAIK GitHub cannot retarget a PR so I will open a new one.

PR #68, opened 2015-05-25
Do not display an error message after aborting YaST (bsc#614829)

lslezak:
Display the main menu directly.

Tested with `sudo ./mksusecd --micro --initrd /tmp/initrd --create /tmp/test.iso openSUSE-13.2-DVD-x86_64.iso`, after aborting the installation the error message is skipped and the main menu is displayed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
 changelog
 Changelog
 VERSION
+*.o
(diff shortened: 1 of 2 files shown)


mchf:
LGTM

PR #70, opened 2015-05-29
- reimplement 'boot installed system' option (bsc #906990)

wfeldt:
- menu is directly reachable via systemboot=1 boot option


shundhammer:
Other than the few minor points I commented, this looks good to me.

PR #71, opened 2015-06-22
- changed 'media not found' message into a dialog letting the user se…

wfeldt:
…lect

  the URL to retry (bsc #923918)

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -1173,19 +1173,14 @@ void lxrc_init()
       config.cd1texts ? "Continue?" : "Choose the URL to retry."
     );
 
-    do {
-      config.manual = 0;
+    if(config.cd1texts) {
+      slist_t *sl;
 
-      if(config.cd1texts) {
-        slist_t *sl;
-
-        if(dia_okcancel(buf, YES) != YES) {
-          config.rescue = 0;
-          config.manual |= 1;
-
-          break;
-        }
-
+      if(dia_okcancel(buf, YES) != YES) {
+        config.rescue = 0;
+        config.manual |= 1;
+      }
+      else {
         for(sl = config.defaultrepo; sl; sl = sl->next) {
           config.manual = 0;
 
@@ -1194,10 +1189,12 @@ void lxrc_init()
 
           if(auto2_find_repo()) break;
         }
-
-        break;
       }
-      else {
+    }
+    else {
+      do {
+        config.manual = 0;
+
         select_repo_url(buf, &repo);
 
         if(!repo || !*repo) {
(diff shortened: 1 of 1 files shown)


linuxrc.c
```diff
@@ -1148,53 +1147,75 @@ void lxrc_init()
...
-    util_disp_init();
+    // ok, we failed to find a suitable repo
+    // do something about it
 
-    i = 0;
-    j = 1;
+    // If the file '/nextmedia' exists, get the message to show from there
+    // and retry the default repo settings if the user is ready.
+    // Otherwise, present the user the list of repos we have tried so far and
+    // let her choose / edit one and try again.
```
jreidinger: :+1: for explaining rationale

linuxrc.c
```diff
@@ -1148,53 +1147,75 @@ void lxrc_init()
...
 
-      strprintf(&buf, "%s\n\nRetry?", s ?: "Please make sure your installation medium is available.");
-      do {
-        j = dia_yesno(buf, YES) == YES ? 1 : 0;
-        if(j) {
-          slist_t *sl;
+    do {
+      config.manual = 0;
+
+      if(config.cd1texts) {
```
jreidinger: work-flow in this if statement is a bit confusing as it actually do not use look from do ( always break at the end ).
So maybe it deserve own method and work-flow like:

```
config.manual = 0;
if(config.cd1texts) {
  ....
} else {
  do {
     ...
  }
}
```
of course unless I overlook that else branch can set `config.cd1texts`,  but from current code it doesn't look like.

linuxrc.c
```diff
@@ -1567,3 +1588,71 @@ void lxrc_readd_parts()
 }
 
 
+/*
+ * Offer the user a list of URLs to choose from based in the current install
+ * URL and the default repo setting.
+ *
+ * In addidtion she can also edit the URL.
```
jreidinger: typo `addidtion`

linuxrc.c
```diff
@@ -1567,3 +1588,71 @@ void lxrc_readd_parts()
...
+    for(i = 1, sl = config.defaultrepo; sl && strcmp(sl->key, config.url.install->str); sl = sl->next) i++;
+    if(sl) default_item = i;
+  }
+
+  i = dia_list(msg, 64, NULL, item_list, default_item, align_left);
+
+  if(i > 0) {
+    if(i == items) {
+      if(!*repo) str_copy(repo, item_list[default_item - 1]);
+      if(dia_input2("Enter the repository URL.", repo, 64, 0)) {
```
jreidinger: is 64 hard limit or soft width of text entry? As I worry that there can be longer URL.

wfeldt: it's just the displayed field size

linuxrc.c
wfeldt: true enough; I'll check the workflow tomorrow, I think it's wrong this way

linuxrc.c
jreidinger: then it is fine, I just want to be sure.

jreidinger:
LGTM

PR #72, opened 2015-06-25
- avoid any interactive dialogs when checking the digest of files during

wfeldt:
an 'extend' command (bsc #936068)


jreidinger:
LGTM

PR #74, opened 2015-07-28
add 'sethostname' option to have wicked set the hostname (bsc #934118)

wfeldt:


wfeldt:
I nearly did but then it would be even more correct to go for uname(2) directly at which point I just thought WTF.

PR #75, opened 2015-08-06
with netsetup=dhcp,all, do network setup at once, else we'll never ge…

wfeldt:
…t it done (bsc #932692)


cwh42:
It's a one-liner. Steffen explained me, what it does. So it LGTM.

PR #76, opened 2015-08-11
Bsc 940465

wfeldt:


markkp:
I guess it looks alright, but as you know I'm not the best person to ask about C coding. ;)  The bugzilla # referenced doesn't seem to have anything to do with this commit, though.

wfeldt:
Oops, that is true. :-(

The correct bugzilla number is bsc #941122.

PR #77, opened 2015-08-31
s390x: treat virtio network device as layer2 device (bsc #943743)

wfeldt:


wfeldt:
see https://github.com/openSUSE/linuxrc/pull/78

PR #80, opened 2015-09-07
s390x: keep ssh active when selecting vnc (bsc #943744)

wfeldt:
revert a bit of the last change


wfeldt:
no

markkp:
So, if linuxrc parses the UseVNC=1, and then UseSSH=1, config.vnc won't be set to 0?

If so, then I'm OK with the change.

wfeldt:
This whole patch has nothing to do with config file parsing; iit's just about that single menu.

PR #84, opened 2015-09-16
get rid of 'no job control' bash messages (bsc #945607)

wfeldt:
Also, make the actual shell command used configurable via debugshell option.


jreidinger:
LGTM

PR #86, opened 2015-10-02
- put udev modprobe rules file into /run so it doesn't get copied int…

wfeldt:
…o the

  final system (bsc #947807)


imobach:
LGTM

PR #87, opened 2015-10-07
ensure ifcfg options don't get lost (bsc #948774)

wfeldt:


net.c
```diff
@@ -2970,3 +2978,56 @@ int compare_subnet(char *ip1, char *ip2, unsigned prefix)
...
+  if(!ifcfg || !device) return;
+
+  ifcfg->flags = slist_free(ifcfg->flags);
+
+  if(config.debug >= 2) fprintf(stderr, "ifcfg flags, before(%s):\n%s", device, ifcfg_print(ifcfg));
+
+  // 1st try, direct match
+  for(tmp = config.ifcfg.all; tmp; tmp = tmp->next) {
+    if(tmp->pattern || !tmp->device) continue;
+    if(config.debug >= 2) fprintf(stderr, "direct:\n%s", ifcfg_print(tmp));
```
jreidinger: just idea, why not have something like debug method that do such if, so in code it can be `debug(2, "direct:\n%s", ifcfg_print(tmp))`

jreidinger: of course disadvantage is expansion of second parameter, but it can be #define macro, so it should not be problem

jreidinger:
just NP, LGTM

net.c
wfeldt: wouldn't improve the situation, imo

There is a long standing plan to have a proper logging function instead of the printf's but that's a bit out of scope here.

PR #88, opened 2015-10-14
set hostname to dhcp provided value per default (bsc #946047)

wfeldt:
After talking to Marius (mt) I chang the default and set the hostname via dhcp.

This will cause the ifcfg-files written by linuxrc to contain DHCLIENT_SET_HOSTNAME=yes.

You can still use the sethostname=0 boot option to revert this.

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -806,6 +806,7 @@ void lxrc_init()
 
   config.ifcfg.manual = calloc(1, sizeof *config.ifcfg.manual);
   config.ifcfg.manual->dhcp = 1;
+  config.net.sethostname = 1;
 
   // config.nanny = 0;	/* see config.nanny_set */
 


imobach:
LGTM

mchf:
it breaks yast2-network's proposal regarding DHCLIENT_SET_HOSTNAME (/etc/sysconfig/network/dhcp is copied to target). Also Marius complains in bugzilla. So NLGTM.

wfeldt:
> Also Marius complains in bugzilla.

???

Marius specifically *asks* me to do it this way.

mchf:
> Marius specifically asks me to do it this way.

ah sorry I've misread the description. I thought that it sets the global flag. 

I assume that linuxrc always configures just one interface, so it places the flag into just one ifcfg. Right?

wfeldt:
No; for every interface the user configures. Normally just one, but if she sets up more, then every ifcfg file gets that attribute.

mchf:
I don't think that setting the flag for every interface is correct, because it makes the behavior a bit unpredictable. 

However, we can postpone this discussion since the fix doesn't solve the problem. The resolver still returns IP address even with this flag set when DHCP server doesn't provide hostname (e.g. "hostname -f" returns IP). In my POV it should fail when the hostname is not resolvable.

wfeldt:
I would actually prefer not to change things here for sle12-sp1 unless there is a regression to sle12.
We're already past rc2.

mchf:
Then we would need to add a hack into yast2-network to fix the issue. Lets ask PO, to choose according his preferences.

kobliha:
@wfeldt Let's close this without merging, hotfix is already implemented...

PR #89, opened 2015-10-15
Do not limit number of inodes in tmpfs overlay

dirkmueller:
On architectures like aarch64 which use 64k pagesize, we're
running very quickly against inode limits otherwise.


dirkmueller:
@wfeldt ping..?

PR #91, opened 2015-11-06
fix wifi support in linuxrc (bsc #949938)

wfeldt:
This is a complete rewrite of the wifi related code to work
with wicked.

linxurc supports open wlan, wpa-psk, and peap authentication.
If you need anything more specific, you can still specify the
required WIRELESS_* parameters via ifcfg option.

There is a new wlan.device option to tag an arbitray interface
as wifi-capable (for testing or to override auto-detection).


dialog.c
```diff
@@ -116,9 +116,8 @@ struct {
   { di_osa_qdio,	 "QDIO"             },
 
   { di_wlan_open,        "No Authentication"   },
-  { di_wlan_wep_o,       "WEP - Open"          },
-  { di_wlan_wep_r,       "WEP - Shared Key"    },
-  { di_wlan_wpa,         "WPA-PSK"             },
+  { di_wlan_wpa_psk,     "WPA-PSK"             },
+  { di_wlan_wpa_peap,    "WPA-PEAP"            },
```
jreidinger: is somewhere documented we support only these two?

global.h
```diff
@@ -367,7 +362,6 @@ typedef struct {
   unsigned startshell:1;	/* start shell before & after yast */
   unsigned listen:1;		/* listen on port */
   unsigned zombies:1;		/* keep zombies around */
-  unsigned forceip:1;		/* do not check for duplicate IP address */
```
jreidinger: how is this related to wifi support?

dialog.c
wfeldt: mentioned in the commit

net.c
```diff
@@ -811,9 +816,9 @@ int net_choose_device()
 
     if(item_hds && item_hds[choice - 1]) {
       hd = item_hds[choice - 1];
-      if(hd->is.wlan) {
-        if(wlan_setup()) choice = -1;
-      }
+//      if(util_is_wlan(config.ifcfg.manual->device)) {
+//        if(wlan_setup()) choice = -1;
+//      }
```
jreidinger: commented out code is smell, please remove it. Git will remember it

global.h
wfeldt: killing a left over, was in an adjacent line

net.c
wfeldt: oops, just forgot to kill it

url.c
```diff
@@ -2506,7 +2506,7 @@ int url_setup_interface(url_t *url)
 
   fprintf(stderr, "interface setup: %s\n", config.ifcfg.manual->device);
 
-  if(url->is.wlan && wlan_setup()) return 0;
+//  if(url->is.wlan && wlan_setup()) return 0;
```
jreidinger: dead code

dialog.c
jreidinger: well, I expect at least some wiki page or something like this. When I try to google linuxrc wifi, results are not much satisfying.

jreidinger:
in general LGTM. would be very nice if googling linuxrc wifi give me something reasonable.

global.h
jreidinger: ok

kobliha:
Oh, huh, this will definitely need a thorough testing according to `documented` parameters. It can't be automated, but we should have all scenarios tested.

PR #92, opened 2015-11-18
rewrite linuxrc logging function

wfeldt:
linuxrc now writes a full log to /var/log/linuxrc.log and a shortend
log to /dev/tty3.

### DIFF
--- a/display.c
+++ b/display.c
@@ -13,6 +13,7 @@
 
 #include "global.h"
 #include "display.h"
+#include "util.h"
 #include "utf8.h"
 
 /*
@@ -208,7 +209,7 @@ void disp_gotoxy(int x, int y)
 {
   if(x == disp_x_im && y == disp_y_im) return;
 
-//  fprintf(stderr, "gotoxy %d x %d\n", x, y);
+//  log_info("gotoxy %d x %d\n", x, y);
 
   if(
     x > 0 && x <= max_x_ig &&
@@ -331,7 +332,7 @@ void disp_save_area(window_t *win)
 
   if(y_len + win->y_left > max_y_ig) y_len = max_y_ig - win->y_left + 1;
 
-//  fprintf(stderr, "save area at %d x %d (size %d x %d)\n", win->x_left, win->y_left, x_len, y_len);
+//  log_info("save area at %d x %d (size %d x %d)\n", win->x_left, win->y_left, x_len, y_len);
 
 //  dump_screen("save area, start");
 
@@ -443,7 +444,7 @@ void disp_restore_screen()
 {
   int x, y;
 
-  fprintf(stderr, "restore screen\n");
+  log_info("restore screen\n");
 
   disp_x_im = 0;
 
@@ -519,7 +520,7 @@ void disp_write_string(char *str)
 {
   int len, *buf;
 
-//  fprintf(stderr, "[* <%s>", str); fflush(stderr);
+//  log_info("[* <%s>", str);
 //  getchar();
 
   len = strlen(str) + 1;
@@ -528,7 +529,7 @@ void disp_write_string(char *str)
   utf8_to_utf32(buf, len, str);
   disp_write_utf32string(buf);
 
-//  fprintf(stderr, "#]\n"); fflush(stderr);
+//  log_info("#]\n");
 
   free(buf);
 }
--- a/file.c
+++ b/file.c
@@ -980,7 +980,16 @@ void file_do_info(file_t *f0, file_key_flag_t flags)
         break;
 
       case key_linuxrcstderr:
-        if(*f->value) util_set_stderr(f->value);
+        if(*f->value) {
+          if(
+            !config.log.dest[1].name ||
+            strcmp(config.log.dest[1].name, f->value)
+          ) {
+            str_copy(&config.log.dest[1].name, f->value);
+            if(config.log.dest[1].f) fclose(config.log.dest[1].f);
+            config.log.dest[1].f = NULL;
+          }
+        }
         break;
 
       case key_kbdtimeout:
@@ -1063,6 +1072,13 @@ void file_do_info(file_t *f0, file_key_flag_t flags)
           }
         }
         slist_free(sl0);
+
+        if(config.error_trace) {
+          config.log.dest[2].level |= LOG_CALLER;
+        }
+        else {
+          config.log.dest[2].level &= ~LOG_CALLER;
+        }
         break;
 
       case key_linuxrc:
@@ -2466,8 +2482,7 @@ void wait_for_conn(int port)
 
   addr.sin_port = htons(port);
 
-  printf("\n%s waiting for connection on port %d\n", inet2print(&config.net.hostname), port);
-  fflush(stdout);
+  log_show("\n%s waiting for connection on port %d\n", inet2print(&config.net.hostname), port);
 
   if(
     (sock = socket(AF_INET, SOCK_STREAM, 0)) == -1 ||
--- a/file.c
+++ b/file.c
@@ -584,7 +584,7 @@ char *file_read_info_file(char *file, file_key_flag_t flags)
   file_t *f0 = NULL;
 
 #ifdef DEBUG_FILE
-  fprintf(stderr, "looking for info file: %s\n", file);
+  log_debug("looking for info file: %s\n", file);
 #endif
 
   if(!strcmp(file, "cmdline")) {
@@ -597,7 +597,7 @@ char *file_read_info_file(char *file, file_key_flag_t flags)
   if(!f0) return NULL;
 
 #ifdef DEBUG_FILE
-  fprintf(stderr, "info file read from \"%s\":\n", file);
+  log_debug("info file read from \"%s\":\n", file);
   file_dump_flist(f0);
 #endif
 
@@ -908,7 +908,7 @@ void file_do_info(file_t *f0, file_key_flag_t flags)
                     char *argv[2] = { };
 
                     argv[1] = buf;
-                    fprintf(stderr, "swapon %s\n", buf);
+                    log_info("swapon %s\n", buf);
                     util_swapon_main(2, argv);
                     break;
                   }
@@ -927,14 +927,14 @@ void file_do_info(file_t *f0, file_key_flag_t flags)
           t = util_fstype(s, NULL);
           if(t && !strcmp(t, "swap")) {
             argv[1] = s;
-            fprintf(stderr, "swapon %s\n", s);
+            log_info("swapon %s\n", s);
             util_swapon_main(2, argv);
           }
         }
         break;
 
       case key_exec:
-        if(*f->value) system(f->value);
+        if(*f->value) lxrc_run_console(f->value);
         break;
 
       case key_usbwait:
@@ -1317,7 +1317,7 @@ void file_do_info(file_t *f0, file_key_flag_t flags)
 	    else {
 	      strprintf(&sl->value, "%s %s", sl->value, s);
 	    }
-            if(config.debug >= 2) fprintf(stderr, "options[%s] = \"%s\"\n", sl->key, sl->value);
+            log_debug("options[%s] = \"%s\"\n", sl->key, sl->value);
           }
         }
         break;
@@ -1824,7 +1824,7 @@ void file_write_install_inf(char *dir)
   strcat(strcpy(file_name, dir), INSTALL_INF_FILE);
 
   if(!(f = fopen(file_name, "w"))) {
-    fprintf(stderr, "Cannot open yast info file\n");
+    log_info("Cannot open yast info file\n");
     return;
   }
 
@@ -1854,7 +1854,7 @@ void file_write_install_inf(char *dir)
   else {
     char buf[256];
     if(!gethostname(buf, sizeof buf)) {
-      if(config.debug) fprintf(stderr, "hostname = \"%s\"\n", buf);
+      log_debug("hostname = \"%s\"\n", buf);
       if(*buf && strcmp(buf, "(none)")) file_write_str(f, key_hostname, buf);
     }
   }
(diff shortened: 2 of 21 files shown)


display.c
```diff
@@ -208,7 +209,7 @@ void disp_gotoxy(int x, int y)
 {
   if(x == disp_x_im && y == disp_y_im) return;
 
-//  fprintf(stderr, "gotoxy %d x %d\n", x, y);
+//  log_info("gotoxy %d x %d\n", x, y);
```
jreidinger: commented out code. Looks useless for me

display.c
```diff
@@ -331,7 +332,7 @@ void disp_save_area(window_t *win)
 
   if(y_len + win->y_left > max_y_ig) y_len = max_y_ig - win->y_left + 1;
 
-//  fprintf(stderr, "save area at %d x %d (size %d x %d)\n", win->x_left, win->y_left, x_len, y_len);
+//  log_info("save area at %d x %d (size %d x %d)\n", win->x_left, win->y_left, x_len, y_len);
```
jreidinger: same here

display.c
```diff
@@ -519,7 +520,7 @@ void disp_write_string(char *str)
 {
   int len, *buf;
 
-//  fprintf(stderr, "[* <%s>", str); fflush(stderr);
+//  log_info("[* <%s>", str);
 //  getchar();
```
jreidinger: and here

display.c
```diff
@@ -528,7 +529,7 @@ void disp_write_string(char *str)
   utf8_to_utf32(buf, len, str);
   disp_write_utf32string(buf);
 
-//  fprintf(stderr, "#]\n"); fflush(stderr);
+//  log_info("#]\n");
```
jreidinger: and here

file.c
```diff
@@ -584,7 +584,7 @@ char *file_read_info_file(char *file, file_key_flag_t flags)
   file_t *f0 = NULL;
 
 #ifdef DEBUG_FILE
-  fprintf(stderr, "looking for info file: %s\n", file);
+  log_debug("looking for info file: %s\n", file);
```
jreidinger: I do not get much why it needs to be closed in DEBUG_FILE directive? Is it some kind of mentioned info file?

global.h
```diff
@@ -94,6 +94,39 @@ enum langid_t {
 
 #define LXRC_WAIT util_wait(__FILE__, __LINE__, __FUNCTION__);
 
+// LOG_* are bitmasks
+#define LOG_LEVEL_SHOW	(1 << 0)
+#define LOG_LEVEL_INFO	(1 << 1)
+#define LOG_LEVEL_DEBUG	(1 << 2)
```
jreidinger: looks likr you would like to have indented values and this line break it ( at least when viewing on github )

global.h
```diff
@@ -94,6 +94,39 @@ enum langid_t {
...
+
+// log to the default console
+#define log_show(...) util_log(LOG_LEVEL_SHOW, __VA_ARGS__)
+// log to the logging console
+#define log_info(...) util_log(LOG_LEVEL_INFO, __VA_ARGS__)
+// log to the log file
+#define log_debug(...) util_log(LOG_LEVEL_DEBUG, __VA_ARGS__)
+// log to the default console if cond is true
+// else log to logging console
+#define log_show_maybe(cond, ...) util_log((cond) ? LOG_LEVEL_SHOW : LOG_LEVEL_INFO, __VA_ARGS__)
```
jreidinger: I am still thinking about better name as I do not like much `maybe` It is also a bit confusing as intuitive replacement `conditionally` indicate that it do not log anything if failed. maybe something like `log_show_or_info` ?

global.h
```diff
@@ -94,6 +94,39 @@ enum langid_t {
...
+// log to the logging console if cond is true
+// else log only to log file
+#define log_info_maybe(cond, ...) util_log((cond) ? LOG_LEVEL_INFO : LOG_LEVEL_DEBUG, __VA_ARGS__)
```
jreidinger: same here.

global.h
jreidinger: btw one advantage of suggested name is also that it shows second log level, which is otherwise hidden in implementation or documentation.

install.c
```diff
@@ -1219,28 +1217,28 @@ int inst_execute_yast()
       inst_pid = fork();
 
       if(inst_pid) {
-        // fprintf(stderr, "%d: inst_pid = %d\n", getpid(), inst_pid);
+        // log_info("%d: inst_pid = %d\n", getpid(), inst_pid);
```
jreidinger: commented out code and much more below.

keyboard.c
```diff
@@ -202,7 +203,7 @@ int kbd_getch(int wait_iv)
   do {
     key = kbd_getch_raw(wait_iv);
 
-//    fprintf(stderr, "rawkey(%d): %d\n", wait_iv, key);
+//    log_info("rawkey(%d): %d\n", wait_iv, key);
```
jreidinger: commented out code.

keyboard.c
```diff
@@ -290,7 +291,7 @@ void check_for_key(int del_garbage)
       kbd.key = utf8_decode(kbd.buffer);
       del_keys(u);
 
-//      fprintf(stderr, "utf8(%u) = 0x%02x\n", u, kbd.key);
+//      log_debug("utf8(%u) = 0x%02x\n", u, kbd.key);
```
jreidinger: same here and above

linuxrc.c
```diff
@@ -177,11 +192,12 @@ int main(int argc, char **argv, char **env)
       find_shell();
       mount("proc", "/proc", "proc", 0, 0);
       file_do_info(file_get_cmdline(key_tmpfs), kf_cmd + kf_cmd_early);
+      file_do_info(file_get_cmdline(key_linuxrcstderr), kf_cmd + kf_cmd_early);
       file_do_info(file_get_cmdline(key_lxrcdebug), kf_cmd + kf_cmd_early);
       util_free_mem();
       umount("/proc");
 
-      // fprintf(stderr, "free: %d, %d\n", config.memory.free, config.tmpfs);
+      // log_info("free: %d, %d\n", config.memory.free, config.tmpfs);
```
jreidinger: commented out code

display.c
wfeldt: Al those are helpful debug lines, if needed.

net.c
```diff
@@ -2540,13 +2530,11 @@ void net_update_state()
...
   }
 
-  if(config.debug) {
-    fprintf(stderr, "net_update_state: ");
-    for(sl = config.ifcfg.if_state; sl; sl = sl->next) {
-      fprintf(stderr, "%s: %s%s", sl->key, sl->value, sl->next ? ", " : "");
-    }
-    fprintf(stderr, "\n");
+  log_debug("net_update_state: ");
+  for(sl = config.ifcfg.if_state; sl; sl = sl->next) {
```
jreidinger: well, not exactly same code as it iterate even when debug is disabled, but I do not expect performance change.

file.c
wfeldt: It debugs file loading and parsing. You normally don't do this and don't want that logged.

url.c
```diff
@@ -87,7 +87,7 @@ void url_read(url_data_t *url_data)
   digest_init(url_data);
 
   c_handle = curl_easy_init();
-  // fprintf(stderr, "curl handle = %p\n", c_handle);
+  // log_info("curl handle = %p\n", c_handle);
```
jreidinger: commented out code

url.c
```diff
@@ -650,7 +649,7 @@ char *url_print(url_t *url, int format)
   static char *buf = NULL, *s;
   int q = 0;
 
-  // printf("start buf = %p\n", buf);
+  // log_info("start buf = %p\n", buf);
```
jreidinger: again ot of commented out code.

global.h
wfeldt: I actually considered this and found it more confusing.

keyboard.c
wfeldt: You don't need to point out every such line.

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
 
+
+/*
+ * Write log message.
+ *
+ * level is a bitmask determining the destination to log to.
+ *
+ * Add a time stamp when the destination has the LOG_TIMESTAMP flag set.
+ */
+void util_log(unsigned level, char *format, ...)
```
jreidinger: well, one idea. I see that very often ( >90% ) format contains `\n` at the end. I also find it quite common to have endline at the end of log message. Do you consider adding it by default there? Only case I see from this review is something like `checking....` and then `done` which can be done on next line with something like `checking finished`

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+
+  time_t t = time(NULL);
+  struct tm *gm = gmtime(&t);
+
+  va_start(args, format);
+  if(vasprintf(&buf, format, args) == -1) buf = NULL;
+  if(buf) buf_len = strlen(buf);
+  va_end(args);
+
+  for(lf = config.log.dest; lf < config.log.dest + sizeof config.log.dest / sizeof *config.log.dest; lf++) {
```
jreidinger: is possible to precount of condition like 
```
log_file_t lf_end = config.log.dest + sizeof config.log.dest / sizeof * config.log.dest;
for(lf = config.log.dest; lf < lf_end; lf++)
```
that change make readability of for cycle much better.

util.c
wfeldt: Yes, I considered. But for logging to the console I want to keep the option not to have a newline.

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+    if((level & lf->level)) {
```
jreidinger: trick I learned from rubocop. using quit exits that do not increase indentation:
```
if(!(level & lf->level))
  continue;

...
```

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+      if(!lf->f && lf->name) {
+        lf->f = fopen(lf->name, "a");
+      }
+      if(lf->f) {
```
jreidinger: if open do not succeed then maybe I consider using break here like
```
if(!(lf->f))
  break;
```

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+        if((lf->level & LOG_TIMESTAMP)) {
+          if(gm) {
```
jreidinger: can gmtime failed at all? I know that manpage allow it, so maybe we should also documented that we cannot get time? just idea

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+            fprintf(lf->f, "%02d:%02d:%02d <%u>", gm->tm_hour, gm->tm_min, gm->tm_sec, level);
```
jreidinger: so level is print only with timestamp? Probably make sense for outputing to stderr.

util.c
wfeldt: I'm aware of that and done this numerous times in the linuxrc code. So you might also have learned from me... ;-)

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+            if((lf->level & LOG_CALLER)) {
+              if(!caller) caller = util_get_caller(1);
+              if(caller) fprintf(lf->f, " %-28s", caller);
```
jreidinger: nice shortening :+1:

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+            }
+            fprintf(lf->f, ": ");
+          }
+        }
+        if(buf_len) {
```
jreidinger: does it make sense to print timestamp and then not write anything?

util.c
wfeldt: that's not the same

util.c
wfeldt: level is actually a bitmask

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+          fwrite(buf, buf_len, 1, lf->f);
+          // supplement a missing trailing newline
+          if(
+            buf[buf_len - 1] != '\n' &&
+            (lf->level & LOG_TIMESTAMP)
+          ) {
+            fputc('\n', lf->f);
```
jreidinger: so it automatic add newline if printed with timestamp...quite confusing for me...why not if there is caller only?

jreidinger: ah, now I see logic, caller without timestamp is not possible.

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+
+
+/*
+ * Run command and redirect and log stderr to linuxrc log file.
+ *
+ * If log_stdout is != 0, redirect and log also stdout.
+ */
+int util_run(char *cmd, unsigned log_stdout)
+{
+  char buf[10], *buf2 = NULL, *cmd2 = NULL;
```
jreidinger: is buffer size 10 enough? I expect at least 512 or 2048 here.

util.c
wfeldt: Like a blank line (or newline). Yes.

The log function was called, so we log.

util.c
wfeldt: Yes, and I want the log file properly formatted. So it auto-adds newlines. The 'wait ... ok' scheme doesn't work here anyway due to the added time stamp.

util.c
```diff
@@ -5322,3 +5286,159 @@ int util_is_wlan(char *device)
...
+
+  fd = open("/tmp", O_TMPFILE | O_RDWR, S_IRUSR | S_IWUSR);
+
+  if(fd < 0) return err;
+
+  strprintf(&cmd2, "%s 2>&%d%s", cmd, fd, log_stdout ? " >&2" : "");
+
+  err = WEXITSTATUS(system(cmd2));
+
+  log_info_maybe(config.debug, "exec: %s = %d\n", cmd, err);
```
jreidinger: this is quite confusing for me. Why condition is `config.debug`? I expect that if you use `log_debug` and `config.debug` is enabled, it will be printed to same as if it is info.

util.c
wfeldt: Oops, left over from testing partial reads. Adjusted.

jreidinger:
in general it improves logging, just few parts are a bit confusing for code reader.

util.c
wfeldt: I don't want every command that is run logged on the console. But you can have it with linuxrc.debug=1 if you need.

The point is to keep the debugging _console_ a bit reduced, so that the info is not scrolled away too easily.

util.c
jreidinger: ah, now I see. There can be more loggers and if some of them failed to write, then continue with next one. You are right

util.c
jreidinger: yeah, it is also possible :)

global.h
jreidinger: OK, I cannot find easily better name, so lets keep this one.

PR #93, opened 2016-01-22
fix routing config when using vlan (bsc #961624)

wfeldt:


net.c
```diff
@@ -2229,11 +2247,11 @@ int _ifcfg_write(char *device, ifcfg_t *ifcfg)
         // set explicit route to gw unless gw is in the same ipv4 subnet
         // note: we might as well set it always
         if(!compare_subnet(v4_ip, sl1->key, v4_prefix)) {
-          strprintf(&sl->key, "%s - - %s", sl1->key, device);
+          strprintf(&sl->key, "%s - - %s%s", sl1->key, device, vlan ?: "");
```
jreidinger: above it is device *dot* vlan and here it is without dot, intention?

jreidinger:
just one question, otherwise LGTM

net.c
wfeldt: no, dot is included

PR #95, opened 2016-01-29
retry loop.ko module loading (bsc #964046)

wfeldt:
Apparently it can happen that the kernel fails to load a module if it's
'too early'.

As a workaround, we retry a few times. This seems to help...

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -61,6 +61,9 @@
 #define MNT_DETACH	(1 << 1)
 #endif
 
+// maximum number of times we try to load the loop module
+#define MAX_LOOP_TRIES	10
+
 static void lxrc_main_menu     (void);
 static void lxrc_catch_signal_11(SIGNAL_ARGS);
 static void lxrc_catch_signal  (int signum);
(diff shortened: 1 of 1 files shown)


linuxrc.c
```diff
@@ -1558,7 +1561,15 @@ void lxrc_add_parts()
...
         insmod_done = 1;
-        lxrc_run("/sbin/insmod /modules/loop.ko max_loop=64");
+        for(i = 0; i < MAX_LOOP_TRIES; i++) {
+          if(!lxrc_run("/sbin/insmod /modules/loop.ko max_loop=64")) break;
+          sleep(1);
+        }
+        if(i == MAX_LOOP_TRIES) {
+          log_show("Hm, module loading seems not to work, giving up.\n");
+          config.debugwait = 1;
+          LXRC_WAIT
```
mvidner: is a semicolon missing after LXRC_WAIT?

mvidner:
```c
#define LXRC_WAIT util_wait(__FILE__, __LINE__, __FUNCTION__);
```
So it isn't.

mvidner:
LGTM.

wfeldt:
It actually failed due to a glibc bug. It was fixed, so this patch is not required.

PR #96, opened 2016-02-02
workaround: include O_DIRECTORY so open(, O_TMPFILE) works on ppc64 (…

wfeldt:
[bsc#964709](https://bugzilla.suse.com/show_bug.cgi?id=964709)

### DIFF
--- a/util.c
+++ b/util.c
@@ -5356,9 +5356,14 @@ int util_run(char *cmd, unsigned log_stdout)
 
   if(!cmd) return err;
 
-  fd = open("/tmp", O_TMPFILE | O_RDWR, S_IRUSR | S_IWUSR);
+  // workaround: include O_DIRECTORY so it works on ppc64 (bsc #964709)
+  fd = open("/tmp", O_TMPFILE | O_DIRECTORY | O_RDWR, S_IRUSR | S_IWUSR);
 
-  if(fd < 0) return err;
+  if(fd == -1) {
+    perror_debug("failed to create tmp file");
+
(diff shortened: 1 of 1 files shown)


wfeldt:
See the mentioned bug; the open() call fails because the kernel thinks O_DIRECTORY is not set.

markkp:
I did look at the bug.  But for someone looking at the just the code, or seeing the error message, is the error truly that a file was not created, or that the /tmp directory was not opened successfully?

mvidner:
@markkp I believe it is this test that fails: https://github.com/torvalds/linux/blob/cc673757e24d018d64ff8038e28835db1e2902c4/fs/open.c#L910

markkp:
@mvidner That doesn't answer my question of what the linuxrc code itself is trying to do.  At that point in the code are we trying to open a file, or a directory?  If it's the /tmp directory, the error message should say that. If it's actually a file named /tmp, then leave things as they are.

wfeldt:
Please, look at the bug. Sample code is https://bugzilla.suse.com/show_bug.cgi?id=964709#c3

mvidner:
@markkp We are trying to open a file: `open(path, O_TEMPFILE...)` is documented to mean "open an unnamed **file** in the filesystem of the **directory** *path*. http://man7.org/linux/man-pages/man2/open.2.html

mvidner:
IMHO this is not the proper place to *fix* it, but as a *workaround* it is fine.
LGTM.

PR #97, opened 2016-02-05
allow spaces in nameserver option for compatibility (bsc #965293)

wfeldt:


wfeldt:
Indeed, the message is wrong.

wfeldt:
https://github.com/openSUSE/linuxrc/pull/98

PR #99, opened 2016-02-11
Workaround O_TMPFILE definition on SLE12 / aarch64 (bsc#965068)

dirkmueller:
The workaround for ppc64 with definiing it including O_DIRECTORY
is not enough, since its actually the bit for the incorrect
O_DIRECTORY that causes it to be rejected as EINVAL by the
kernel for aarch64.


dirkmueller:
(testing still running)

PR #100, opened 2016-02-16
introduce externally handled (user-defined) URL schemes

wfeldt:
To support fate #318899 (hmc file access) a new url scheme would have had to be added to linuxrc.

To make things easier in future, I've introduced a general framework for handling urls in linuxrc - the actual work is delegated to a shell script. It also unifies the existing code a bit.

The whole thing is documented in user_defined_url_schemes.txt.

Note that this even allows to add support for some new url scheme via a driver update.


install.c
```diff
@@ -170,18 +179,27 @@ int inst_menu_cb(dia_item_t di)
...
     di_netsource_http,
     di_netsource_https,
     di_netsource_nfs,
     di_netsource_smb,
     di_netsource_tftp,
-    di_none
   };
 
-  if(!(config.test || config.net.cifs.binary)) items[3] = di_skip;
+ j = 6;        // first 6 items have aleady been filled in
```
shundhammer: Maybe try to avoid magic numbers in the middle of the code? Add some #define ? Or maybe use a sizeof() of a static array?

install.c
```diff
@@ -473,7 +544,7 @@ int inst_choose_partition(char **partition, int swap, char *txt_menu, char *txt_
     items2[item_cnt2++] = strdup(buf);
   }
 
-  if(swap) {
+  if((flags & ICP_SWAP)) {
```
shundhammer: How about adding a test macro since this seems to be used multiple times? TEST_ICP_SWAP() or something like that? Similar with TEST_ICP_BLOCK_ALL()?

install.c
```diff
@@ -509,100 +586,102 @@ int inst_choose_partition(char **partition, int swap, char *txt_menu, char *txt_
...
-          j = dia_contabort(buf, NO);
-          if(j == YES) {
-            sprintf(buf, "/sbin/mkswap %s", dev);
-            if(!lxrc_run(buf)) {
-              log_info("swapon %s\n", dev);
-              if(swapon(dev, 0)) {
-                log_info("swapon: ");
-                perror_info(dev);
-                dia_message("Error activating swap space.", MSGTYPE_ERROR);
+    if((flags & ICP_SWAP)) {
```
shundhammer: In all other cases, the old 'if' condition was if (swap), now it was something completely different - please double-check if this is really what you want

shundhammer:
Aside those minor issues above, this is a tentative LGTM from my side - tentative because I couldn't claim I understood it well enough. Somebody else please have a look at this, too.

install.c
wfeldt: Suggestions welcome. A define or sizeof wouldn't do.

install.c
wfeldt: Each test would be used 3 times. I wouldn't go for macros quite yet.

wfeldt:
Related to this: the actual implementation for HMC:
https://github.com/openSUSE/installation-images/pull/86

PR #102, opened 2016-03-08
Add SelfUpdate parameter (FATE#319716)

imobach:
* This parameter enables/disables/configures self-update feature.

### DIFF
--- a/file.c
+++ b/file.c
@@ -1525,6 +1525,7 @@ void file_do_info(file_t *f0, file_key_flag_t flags)
       case key_insecure:
         if(f->is.numeric && f->nvalue) {
           config.secure = 0;
+          config.insecure = 1;
           config.digests.failed = config.sig_failed = 0;
         }
         break;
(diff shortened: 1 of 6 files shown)


wfeldt:
Imo: ok, but it doesn't do anything with the url. Will this change or do you just pass the url? If so,
it might be better not to parse and rebuild the url as this ties you to the linuxrc url schemes.

imobach:
@wfeldt The idea is only to pass the URL as is. I though in using `url_*` functions as some kind of "validator". But thinking it twice, you're right. I'll simplify things a bit. Thanks for the review.

imobach:
I've changed the handling of the URL as you suggested.

I've also added the `Insecure` parameter to `/etc/install.inf` as it's needed by FATE#319716. I though writing `Insecure` as `!config.secure` but I realized that it can be changed [during boot](https://github.com/openSUSE/linuxrc/blob/b904083833ebed372a301d3b544e95134c26fbe2/url.c#L1578) and I don't want that change to affect the self-update feature. For example, you can add your own insecure DUD but sure you'll want the official self-update to be checked.

wfeldt:
I don't understand this insecure/secure thing. Once you skipped one signature check, there's really no point checking any again. You system **has** already been compromised. Or not?

BTW, this is the very reason why linuxrc never accepts anything like insecure=0.

imobach:
@wfeldt Sorry, I've just removed too much :(

wfeldt:
lgtm

imobach:
@wfeldt Thanks!

PR #104, opened 2016-03-10
fix network url handling (fate #318899)

wfeldt:


markkp:
I'm not understanding what this change has to do with the feature for installation from the HMC DVD.

wfeldt:
It's an implementation bug for handling network urls.

kobliha:
Then @markkp is right that you should not reference a HMC feature but a bug describing the problem which is fixed by this PR.

wfeldt:
Ok, see bug 970927.

PR #106, opened 2016-04-25
remove O_TMPFILE workaround needed for bsc #964709 and bsc #965068

wfeldt:
A fixed glibc has been submitted and the workaround is no longer needed.


ancorgs:
lgtm

PR #108, opened 2016-05-04
sanitize hostname handling (fate #319639)

wfeldt:


file.c
```diff
@@ -1895,13 +1895,6 @@ void file_write_install_inf(char *dir)
   if(config.net.realhostname) {
     file_write_str(f, key_hostname, config.net.realhostname);
   }
-  else {
-    char buf[256];
-    if(!gethostname(buf, sizeof buf)) {
-      log_debug("hostname = \"%s\"\n", buf);
-      if(*buf && strcmp(buf, "(none)")) file_write_str(f, key_hostname, buf);
-    }
-  }
```
mchf: What was purpose of this part. Why is it not needed anymore?

mchf:
One question which is caused by lack of knowledge about linuxrc. Otherwise LGTM

file.c
wfeldt: That was added as a temporary solution to pass on the wicked-set hostname and is no longer needed.

mvidner:
LGTM. Thanks for the documentation!

PR #109, opened 2016-05-05
Updated the copyright information that is displayed on the console.

markkp:
Added verbiage to the prompt for the MAC address when a Layer 2 NIC
is being used. In most cases the user shouldn't specify one and take
the default.

This builds fine on all architectures and tested fine on s390x.

### DIFF
--- a/net.c
+++ b/net.c
@@ -1548,7 +1548,10 @@ int net_activate_s390_devs_ex(hd_t* hd, char** device)
       }
       if(config.hwp.layer2 == LAYER2_YES) {
         IFNOTAUTO(config.hwp.osahwaddr) {
-          dia_input2("MAC address", &config.hwp.osahwaddr, 17, 1);
+          dia_info(&win, "Specifying a MAC address is optional.", MSGTYPE_INFO);
+          dia_info(&win, "In most cases letting it default is the correct choice.", MSGTYPE_INFO);
+          dia_info(&win, "Provide one only if you know it is truly necessary.", MSGTYPE_INFO);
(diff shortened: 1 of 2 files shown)


net.c
```diff
@@ -1548,7 +1548,10 @@ int net_activate_s390_devs_ex(hd_t* hd, char** device)
       }
       if(config.hwp.layer2 == LAYER2_YES) {
         IFNOTAUTO(config.hwp.osahwaddr) {
-          dia_input2("MAC address", &config.hwp.osahwaddr, 17, 1);
+          dia_info(&win, "Specifying a MAC address is optional.", MSGTYPE_INFO);
+          dia_info(&win, "In most cases letting it default is the correct choice.", MSGTYPE_INFO);
+          dia_info(&win, "Provide one only if you know it is truly necessary.", MSGTYPE_INFO);
+          dia_input2("Optional MAC address", &config.hwp.osahwaddr, 17, 1);
```
wfeldt: That would show 4 dialog boxes in a row. Is that what you intended? Wouldn't it look better putting all the texts into a single dia_input2() call (you can use newlines '\n' in the text)?.

markkp: Well, the only place that this shows up is on the z/VM console or the SCLP console, neither of which supports ncurses.  But, I like your idea better.

markkp:
I also changed the call to dia_input2 to change the pw_mode from 1 to 0. Having it as 1 meant that the user doesn't see their input, which was unexpected and confusing.

wfeldt:
lgtm

PR #110, opened 2016-05-18
activate AutoUpgrade if Upgrade + AutoYaST are set (bsc #978252)

wfeldt:
Autoyast looks for an 'AutoUpgrade' option instead of 'Upgrade' for historical reasons. So be nice and turn Upgrade into AutoUpgrade .


schubi2:
LGTM: thanks for the quick fix !

PR #112, opened 2016-06-14
restore standard IO file descriptors when returning from external shell (bsc #983014)

wfeldt:
linuxrc was looping on closed stdin and writing to broken stdout. So, make sure IO works again when we return from sh.

### DIFF
--- a/util.c
+++ b/util.c
@@ -5457,3 +5457,40 @@ void util_set_hostname(char *hostname)
   log_info("set hostname: %s\n", hostname);
 }
 
+
+/*
+ * Run debug shell.
+ *
+ * This takes care of screen settings and restoring standard file
+ * descriptors.
+ */
+void util_run_debugshell()
+{
+  kbd_end(1);
+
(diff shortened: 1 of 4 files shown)


util.c
```diff
@@ -5457,3 +5457,40 @@ void util_set_hostname(char *hostname)
...
+
+/*
+ * Run debug shell.
+ *
+ * This takes care of screen settings and restoring standard file
+ * descriptors.
+ */
+void util_run_debugshell()
+{
+  kbd_end(1);
```
shundhammer: Really kbd_end(1) ? In both code blocks where this is copied from, it's kbd_end(0).

shundhammer:
Other than that one remark (just trying to make sure it wasn't overlooked), LGTM.

util.c
wfeldt: Yes, intentionally. The difference it that it now closes the input descriptor and reopens it later.

shundhammer:
LGTM then.

PR #113, opened 2016-06-22
document XVideo option

wfeldt:


mvidner:
Nice! Please reference https://bugzilla.suse.com/show_bug.cgi?id=974821 in the commit message.

mvidner:
LGTM

PR #114, opened 2016-07-13
fix reading instsys id (related to bsc #974601)

wfeldt:
instsys id reading didn't work due to this typo and so the check for the correct instsys didn't work.


shundhammer:
LGTM

PR #122, opened 2016-10-14
parse netwait option earlier (bsc #995304)

wfeldt:
Obviously it needs to be parsed before we start to configure the network...


imobachgs:
LGTM

PR #124, opened 2016-10-26
add header guards

michal915:


kobliha:
@michal915 Please add some description to the PR. A reason, explanation, anything that says "why", "why now" or "what you expect to get" after this is merged. Thanks!

PR #125, opened 2016-10-31
Disable self-update feature by default

imobachgs:
We've been asked to disable the self-update feature by default in SLE 12 SP2.

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -798,7 +798,8 @@ void lxrc_init()
 
   // defaults for self-update feature
   config.self_update_url = NULL;
-  config.self_update = 1;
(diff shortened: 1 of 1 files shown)


teclator:
There is a bug number or something that could we add to the request? https://bugzilla.suse.com/show_bug.cgi?id=1007034?

fcrozat:
Bug number for this : https://bugzilla.suse.com/show_bug.cgi?id=1001839

teclator:
LGTM

ancorgs:
it would be good to reference the conversation

imobachgs:
I've added the bug number.

imobachgs:
Thanks!

PR #126, opened 2016-11-18
Clean up: removed "regcert", "smturl" and "smtcert" options (bsc#943966)

lslezak:
These options are not used in YaST anymore, removing the dead code.


wfeldt:
lgtm

PR #127, opened 2016-11-22
Implemented Loghost functionality FATE#318971

dwaas:
FATE ENTRY: https://fate.suse.com/318971

The script handling is almost completely done in installation-images

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -1278,6 +1278,8 @@ void lxrc_init()
 
   util_update_kernellog();
 
+  if(config.loghost) util_run_script("remote_log_setup");
+
 #if !(defined(__PPC__) || defined(__sparc__))
   if(config.manual || reboot_wait_ig) {
     config.rebootmsg = 1;
--- a/util.c
+++ b/util.c
@@ -4549,6 +4549,8 @@ void util_run_script(char *name)
     setenv("linuxrc_debug", buf, 1);
   }
 
+  if(config.loghost) setenv("LOGHOST", config.loghost, 1);
+
   strprintf(&buf, "/scripts/%s", name);
 
   lxrc_run(buf);


linuxrc.c
```diff
@@ -1278,6 +1278,13 @@ void lxrc_init()
 
   util_update_kernellog();
 
+  if(config.loghost)
+  {
+    char cmd[40];
```
mvidner: Only 13 characters are left for the loghost. Consider that even an IP address may be longer than that.

lslezak: 40 bytes buffer seems to be quite a small buffer to me, I guess when user uses a long FQDN it might not fit there. I'd use at least 128 byte.

Moreover you should check the `snprintf` result and not run the command when there is not enough space in the buffer (the command is truncated).

mvidner:
Thanks! Please add a reference to the bug or feature this is solving to the commit message.

I wondered whether the right place was chosen for the added code, but I could not decide because `lxrc_init` has over 600 lines. @wfeldt can you give some advice on how to break it up into manageable pieces?

linuxrc.c
```diff
@@ -1278,6 +1278,13 @@ void lxrc_init()
...
+    snprintf(cmd, sizeof(cmd), "/scripts/remote_log_setup %s", config.loghost);
```
wfeldt: please use asprintf() to avoid the size restrictions

dwaas: I agree that the length of the string should be increased. However I'd prefer allocating in the stack with something like `char cmd[100];`. Any reason to prefer the dynamic allocation?

wfeldt: To avoid the size limit.

wfeldt:
lgtm

PR #130, opened 2016-11-26
shared libraries corner case mentioned in README

dwaas:
Also wrote full working command of makesusecd in the examples.


wfeldt:
lgtm

PR #131, opened 2016-12-14
Write DHCLIENT_SET_HOSTNAME globally (bsc#1013605)

mvidner:
[bsc#1013605](https://bugzilla.suse.com/show_bug.cgi?id=1013605)
https://trello.com/c/pN6f3kzr

### DIFF
--- a/net.c
+++ b/net.c
@@ -77,7 +77,7 @@ static int net_get_ip(char *text, char **ip, int with_prefix);
 static int net_check_ip(char *buf, int multi, int with_prefix);
 static int compare_subnet(char *ip1, char *ip2, unsigned prefix);
 static void get_and_copy_ifcfg_flags(ifcfg_t *ifcfg, char *device);
-static void update_sysconfig(slist_t *slist, char *filename, char *filename_tmp);
+static void update_sysconfig(slist_t *slist, char *filename);
 
 
 /*
@@ -2282,11 +2282,8 @@ int _ifcfg_write(char *device, ifcfg_t *ifcfg)
   }
 
   // 5. update global network config
-
   if(sl_global) {
-    update_sysconfig(sl_global,
-                     "/etc/sysconfig/network/config",
-                     "/etc/sysconfig/network/config.tmp");
+    update_sysconfig(sl_global, "/etc/sysconfig/network/config");
   }
 
   ok = 1;
@@ -2308,10 +2305,11 @@ err:
 
 // Update the sysconfig file *filename* with the data in *slist*.
 //
-// slist is mutated by appending "=" to the keys
-void update_sysconfig(slist_t *slist, char *filename, char *filename_tmp) {
+// *slist* is mutated by appending "=" to the keys.
+// A temporary file (*filename*.tmp) is used.
+void update_sysconfig(slist_t *slist, char *filename) {
   slist_t *sl;
-  FILE *fp, *fp2;
+  FILE *fp, *fp2 = NULL;
 
   log_info("adjusting %s:\n", filename);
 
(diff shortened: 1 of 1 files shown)


mvidner:
Er, I haven't tested this yet. Please tell me whether the approach makes sense first.

net.c
```diff
@@ -2339,6 +2306,52 @@ int _ifcfg_write(char *device, ifcfg_t *ifcfg)
...
+  if((fp = fopen(filename, "r"))) {
+    char buf[4096];
+
+    // we allow open to fail and check fp2 for NULL later
+    fp2 = fopen(filename_tmp, "w");
+
+    while(fgets(buf, sizeof buf, fp)) {
+      if(*buf && *buf != '#' && !isspace(*buf)) {
+        for(sl = slist; sl; sl = sl->next) {
+          if(!strncmp(buf, sl->key, strlen(sl->key))) {
```
mchf: i think that this comparison can corrupt values when a file is written in not so expected order. Lets have ifcfg file:
```
IPADDR_0=1.1.1.1
IPADDR=2.2.2.2
```
and ```slist``` with key: IPADDR and value 3.3.3.3. Then updating the file should end with
```
IPADDR_0=3.3.3.3
IPADDR=2.2.2.2
```
which is incorrect.

mchf: ok, i'll answer myself ... that is why is '=' appended a few lines above ;-)

net.c
```diff
@@ -2339,6 +2306,52 @@ int _ifcfg_write(char *device, ifcfg_t *ifcfg)
   return ok;
 }
 
+// update the sysconfig file *filename* with the data in *slist*
+//
+// slist is mutated by appending "=" to the keys
```
mchf: it would at least deserve mentioning that the procedure requires usage of a writable temporary file

net.c
```diff
@@ -2152,7 +2153,7 @@ int _ifcfg_write(char *device, ifcfg_t *ifcfg)
 
   // set hostname, if requested
   if(config.net.sethostname) {
```
mchf: so, it will work if I use ```SetHostname=1``` as bootparam. But I don't see if this flag is set in case of ```ifcfg=*=dhcp``` too

mvidner: Good point. I think that it *somehow* works, because this was the only place in the source that mentions [`DHCLIENT_SET_HOSTNAME`](https://github.com/openSUSE/linuxrc/search?utf8=%E2%9C%93&q=DHCLIENT_SET_HOSTNAME) ... ... ... and the answer is: [lxrc_init](https://github.com/openSUSE/linuxrc/blob/3b0c3621093304fd5319ca3edb4f4338a94b3156/linuxrc.c#L780)

net.c
```diff
@@ -2322,9 +2319,12 @@ void update_sysconfig(slist_t *slist, char *filename, char *filename_tmp) {
 
   if((fp = fopen(filename, "r"))) {
     char buf[4096];
+    char *filename_tmp;
```
wfeldt: Don't forget to free() the string at the end.

net.c
```diff
@@ -2018,6 +2018,7 @@ int _ifcfg_write(char *device, ifcfg_t *ifcfg)
   slist_t *sl_ifcfg = NULL;
   slist_t *sl_ifroute = NULL;
   slist_t *sl_global = NULL;
+  slist_t *sl_dhcp = NULL;
```
wfeldt: List must be explicitly freed() at function end; cf. sl_global;

mvidner: Duh, of course. Fixed, along with `filename_tmp`.

net.c
```diff
@@ -2349,6 +2350,8 @@ void update_sysconfig(slist_t *slist, char *filename, char *filename_tmp) {
     else {
       log_info("warning: %s not updated\n", filename);
     }
+
+    free(filename_tmp);
```
wfeldt: asprintf() may leave filename_tmp undefined on error.

C is a pain sometimes. :-)

mvidner: Ah, thanks, another thing that I saw in your code and then forgot to apply :-/

mchf:
lgtm (from local -> global setup POV)

wfeldt:
lgtm now; did you actually test this?

mvidner:
Now I have tested it:
- made an RPM out of this with `osc build` for Factory
- built an ISO with `mksusecd` on top of Tumbleweed-DVD-x86_64-Snapshot20161214
- booted with `ifcfg=*=dhcp`
- `grep HOSTNAME /etc/sysconfig/network/*`

Without this fix, `DHCLIENT_SET_HOSTNAME` is `no` in `dhcp` and `yes` in `ifcfg-ens3`, with this fix, it is `yes` in `dhcp`, as it should.

PR #132, opened 2016-12-21
add linuxrc.core option to enable core dumps in linuxrc (to debug bsc #1010505)

wfeldt:
It's used with either block or char dev as arg: linuxrc.core=block_dev|char_dev.
For a block dev, the device is mounted and core files created there. For a char dev,
the uuencoded core is written to it (think of serial line).


shundhammer:
LGTM

PR #133, opened 2017-01-13
have linuxrc accept plain repomd repo as install source (fate #322286)

wfeldt:
https://fate.suse.com/322286
linuxrc will check for /content first and when it's missing, check for /repodata/repomd.xml.

### DIFF
--- a/linuxrc_repo.md
+++ b/linuxrc_repo.md
@@ -1,0 +1,95 @@
+#specifying installation repository in linuxrc
+
+##1. regular install media
+
+*(repo meta data are in /suse/setup/descr/)*
+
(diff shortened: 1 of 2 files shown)


linuxrc_repo.md
```diff
@@ -0,0 +1,95 @@
...
+Normally you would use:
+
+```sh
+install=http://download.opensuse.org/tumbleweed/repo/oss
+```
+
+but you can also use repomd:
+
+```sh
+install=http://download.opensuse.org/tumbleweed/repo/oss/suse instsys=../boot/x86_64/root
```
lslezak: This specific repo is actually a mix of SUSE and repomd format. Interesting... :open_mouth:

PR #134, opened 2017-01-18
allow reading of config files from instsys location (/boot/ARCH/) (fate #322283)

wfeldt:
This works by extending the /boot/ARCH/config file syntax (see docs):

Appending '?copy=1' to a component tells linuxrc that this is not an image
to mount but a file to copy to the '/' directoy.

### DIFF
--- a/auto2.c
+++ b/auto2.c
@@ -863,7 +863,7 @@ char *auto2_serial_console()
 void auto2_read_repo_files(url_t *url)
 {
   int i;
-  char *dst = NULL;
+  char *tmp_file = NULL;
   static char *default_list[][2] = {
     { "/media.1/info.txt", "/info.txt" },
     { "/license.tar.gz", "/license.tar.gz" },
@@ -873,10 +873,19 @@ void auto2_read_repo_files(url_t *url)
   };
 
   for(i = 0; i < sizeof default_list / sizeof *default_list; i++) {
-    url_read_file(url, NULL, default_list[i][0], default_list[i][1], NULL, URL_FLAG_NODIGEST);
+    // be careful not to replace an existing file unless we successfully got
+    // a new version
+    str_copy(&tmp_file, new_download());
+    if(
+      !url_read_file(url, NULL, default_list[i][0], tmp_file, NULL, URL_FLAG_NODIGEST + URL_FLAG_OPTIONAL) &&
+      util_check_exist(tmp_file)
+    ) {
+      rename(tmp_file, default_list[i][1]);
+      log_info("mv %s -> %s\n", tmp_file, default_list[i][1]);
+    }
   }
 
-  str_copy(&dst, NULL);
+  str_copy(&tmp_file, NULL);
 
   if(!config.autoyast) {
     if(util_check_exist("/tmp/autoinst.xml")) rename("/tmp/autoinst.xml", "/autoinst.xml");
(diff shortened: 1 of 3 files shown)


auto2.c
```diff
@@ -873,10 +873,19 @@ void auto2_read_repo_files(url_t *url)
   };
 
   for(i = 0; i < sizeof default_list / sizeof *default_list; i++) {
-    url_read_file(url, NULL, default_list[i][0], default_list[i][1], NULL, URL_FLAG_NODIGEST);
+    // be careful not to replace an existing file unless we successfully got
+    // a new version
+    str_copy(&tmp_file, new_download());
+    if(
+      !url_read_file(url, NULL, default_list[i][0], tmp_file, NULL, URL_FLAG_NODIGEST + URL_FLAG_OPTIONAL) &&
+      util_check_exist(tmp_file)
```
shundhammer: Why first try to read the file and only if that fails check if it exists? Wouldn't it make more sense the other way round?

url.c
```diff
@@ -2269,19 +2270,30 @@ static int test_is_repo(url_t *url)
...
-        log_info("mount %s -> %s\n", buf, sl->value);
-
-        i = util_mount_ro(buf, sl->value, url->file_list) ? 0 : 1;
-        ok &= i;
-        if(!i) log_info("instsys mount failed: %s\n", sl->value);
+        if(copy) {
+          char *argv[3];
+          char *dst = strrchr(t, '/') ?: t;
+          log_info("copy %s -> %s\n", buf, dst);
+          argv[1] = buf;
```
shundhammer: What about argv[0]? This remains uninitialized here.

url.c
```diff
@@ -2301,11 +2313,19 @@ static int test_is_repo(url_t *url)
         buf2,
         URL_FLAG_PROGRESS + URL_FLAG_UNZIP + opt * URL_FLAG_OPTIONAL
       )) {
-        log_info("mount %s -> %s\n", file_name, sl->value);
-
-        i = util_mount_ro(file_name, sl->value, url->file_list) ? 0 : 1;
-        ok &= i;
-        if(!i) log_info("instsys mount failed: %s\n", sl->value);
+        if(copy) {
+          char *dst = strrchr(t, '/') ?: t;
```
shundhammer: Looks like one parameter of the ? operator is missing - why does this even compile?

auto2.c
wfeldt: The check will always fail before we read the file, or?

url.c
wfeldt: argv[0] would normally be the command name; the field is never read

url.c
wfeldt: valid C syntax; a ?: x is short for a ? a :  x, but a is evaluated only once

url.c
shundhammer: Right, but better be safe than sorry; leaving it with memory garbage is not a good idea IMHO. I'd at least initialize it with NULL to get a honest core dump rather than random results when passing memory garbage.

url.c
shundhammer: OK... interesting. I hadn't known that.

auto2.c
shundhammer: Ah, it checks for the other file... ok, stupid me.

url.c
aschnell: GNU C accorinding to https://en.wikipedia.org/wiki/%3F:#C.

shundhammer: Arvin just found out that this is a GNU extension. But that's good enough for us since we can rely on having GNU C.

shundhammer:
LGTM

shundhammer:
LGTM now.

PR #135, opened 2017-01-19
allow reading of config files from instsys location, part 2

wfeldt:
It was necessary to adjust the code in a second place.


shundhammer:
LGTM

PR #139, opened 2017-04-27
read license.tar.gz from repomd repositories (fate #322286)

wfeldt:
This includes downloading and parsing repomd.xml to find ot the real file name
and file checksum.

This reuses the existing **very** limited xml parser file_parse_xmllike() which
seems sufficient here.

### DIFF
--- a/auto2.c
+++ b/auto2.c
@@ -925,7 +925,7 @@ void auto2_read_repomd_files(url_t *url)
     // be careful not to replace an existing file unless we successfully got
     // a new version
 
-    // first, get real file name
+    // get real file name
     slist_t *sl = slist_getentry(config.repomd_data, default_list[i][0]);
     if(!sl) continue;
 
--- a/auto2.c
+++ b/auto2.c
@@ -37,6 +37,7 @@
 static int driver_is_active(hd_t *hd);
 static void auto2_progress(char *pos, char *msg);
 static void auto2_read_repo_files(url_t *url);
+static void auto2_read_repomd_files(url_t *url);
 static char *auto2_splash_name(void);
 static void auto2_kexec(url_t *url);
 
@@ -601,7 +602,10 @@ int auto2_find_repo()
   }
 
   /* get some files for lazy yast */
-  if(!err) auto2_read_repo_files(config.url.install);
+  if(!err) {
+    if(config.repomd) auto2_read_repomd_files(config.url.install);
+    auto2_read_repo_files(config.url.install);
+  }
 
   if(err) {
     log_info("no %s repository found\n", config.product);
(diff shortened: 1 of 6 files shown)


auto2.c
```diff
@@ -601,7 +602,10 @@ int auto2_find_repo()
   }
 
   /* get some files for lazy yast */
-  if(!err) auto2_read_repo_files(config.url.install);
+  if(!err) {
+    if(config.repomd) auto2_read_repomd_files(config.url.install);
+    auto2_read_repo_files(config.url.install);
```
jreidinger: so if config.repodm is set, then both reads are called? is it intention? or only read_rpomd should be called?

auto2.c
```diff
@@ -905,6 +909,41 @@ void auto2_read_repo_files(url_t *url)
...
+  char *tmp_file = NULL;
+  static char *default_list[][2] = {
+    { "license", "/license.tar.gz" },
+  };
+
+  for(i = 0; i < sizeof default_list / sizeof *default_list; i++) {
+    // be careful not to replace an existing file unless we successfully got
+    // a new version
+
+    // first, get real file name
```
jreidinger: if it is only comment, I suggest to remove "first" as I try to search for second which I do not see.

file.c
```diff
@@ -2618,3 +2616,83 @@ slist_t *file_parse_xmllike(char *name, char *tag)
...
+  if(!tag) return sl0;
+
+  if(!(f = fopen(name, "r"))) return sl0;
+
+  do {
+    buf = realloc(buf, buf_size += 0x1000);
+    i = fread(buf + buf_ptr, 1, buf_size - buf_ptr - 1, f);
+    buf_ptr += i;
+  }
+  while(buf_ptr == buf_size - 1);
```
jreidinger: is this style unified accross files? I expect to have "while" after `}` as now it looks for me like it is only `while(buf_ptr == buf_size - 1);` which is valid, just never ends. but if it is consistent with rest of code, then fine for me

file.c
```diff
@@ -2618,3 +2616,83 @@ slist_t *file_parse_xmllike(char *name, char *tag)
...
+
+  buf[buf_ptr] = 0;
+
+  fclose(f);
+
+  if(!(buf_size = buf_ptr)) return sl0;
```
jreidinger: is `=` intetion? shouldn't it be `==` for comparison?

auto2.c
wfeldt: Intentionally; it reads e.g. the autoinst.xml file this way.

jreidinger: ok, thanks for info. Can you put it also to comment in code?

file.c
wfeldt: Consistent, also same for if ... else.

file.c
wfeldt: Good catch.

(a) yes, '=' was intended.
(b) after the splitting of the function it's no longer needed at all

PR #140, opened 2017-05-19
enforce linuxrc's maximum screen dimensions properly (bsc #1027354)

wfeldt:
linuxrc uses static arrays for its textmode dialogs. The boundary check for
too large screens exsisted but was not always employed.


ancorgs:
LGTM

PR #146, opened 2017-09-20
pass SetHostname setting to install.inf

wfeldt:


teclator:
This PR is part of this trello card:

https://trello.com/c/Vj69gtYP/1168-5-caasp-20-hostname-via-dhcp-work-according-to-control-file

Basically we need to know if the user already disabled it by linuxrc before preload defaults from the control file.

teclator:
LGTM

PR #147, opened 2017-09-25
add SetHostnameUsed entry to install.inf (bsc #1054933)

wfeldt:
SetHostnameUsed is set if the user has used the SetHostname option
explicitly to change its value.

Note that this not only means via kernel command line but also includes
config files like linuxrc.config or an info file.

### DIFF
--- a/global.h
+++ b/global.h
@@ -633,6 +633,7 @@ typedef struct {
     unsigned ipv6:1;		/**< do ipv6 config */
     unsigned dhcp_timeout_set:1;	/**< dhcp_timeout was set explicitly */
     unsigned sethostname:1;	/**< wicked should set hostname */
+    unsigned sethostname_used:1;	/**< user has used linuxrc's SetHostname option */
     unsigned do_setup;		/**< do network setup */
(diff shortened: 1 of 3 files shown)


global.h
```diff
@@ -633,6 +633,7 @@ typedef struct {
     unsigned ipv6:1;		/**< do ipv6 config */
     unsigned dhcp_timeout_set:1;	/**< dhcp_timeout was set explicitly */
     unsigned sethostname:1;	/**< wicked should set hostname */
+    unsigned sethostname_used:1;	/**< user has used linuxrc's SetHostname option */
```
teclator: Is it basically always set?

teclator:
It is related to:

https://github.com/openSUSE/linuxrc/pull/146

And also part of the same [PBI](https://trello.com/c/Vj69gtYP/1168-5-caasp-20-hostname-via-dhcp-work-according-to-control-file)

We are not only interested in the SetHostname value but particulary if it was established by the user.

teclator:
LGTM

PR #151, opened 2017-10-23
Support for "media_upgrade" boot option (fate#323163), enabled Travis, added smoke test

lslezak:
### Summary

- Added a new handler for the `media_upgrade` boot option ([fate#323163](https://fate.suse.com/323163))
- Added a basic Travis support (does not build the RPM, just compiles the sources), but at least something...
- The `git2log` script has been updated to handle detached Git HEAD in Travis build (just updated from the [linuxrc-devtools](https://github.com/openSUSE/linuxrc-devtools) repository)

### Travis Smoke Test

Added a simple smoke test to ensure `linuxrc` can be started and does not crash.

  - The test covers both the line mode and the interactive mode
  -  See an [example Travis log](https://travis-ci.org/openSUSE/linuxrc/builds/291400319#L975) for the result
  - Unfortunately Travis does not display the screen dump well, but it is still somehow readable (the colors are bit broken, too big line spacing, the output contains some broken UTF-8 characters), when running locally it looks correctly
  - Also I could not find why sending the `Ctrl+C` keyboard combination in Travis does not work, when running the Docker build locally it works. So there is a small hack to finish the linuxrc interactive test earlier when `TRAVIS` environment variable is set.

### Testing 

Tested manually using the `mksusecd` script, when booted with `upgrade=1 media_upgrade=1` then the `/etc/install.inf` contains the expected values:

linuxrc_media_upgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+sudo: required
+language: bash
+services:
+  - docker
+
+before_install:
+  - docker build -t linuxrc-image .
+  # list the installed packages (just for easier debugging)
+  - docker run --rm -it linuxrc-image rpm -qa | sort
+
+script:
+  - docker run --rm -it -e TRAVIS=1 linuxrc-image bash -c "make -j `nproc` && ./smoke_test.sh"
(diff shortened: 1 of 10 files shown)


wfeldt:
lgtm; although I think running just in linemode would be enough in travis.

lslezak:
I'll keep it, it also tests the UI related code a bit. If it makes troubles at Travis we can remove it later...

PR #154, opened 2017-11-02
Update makefile

plusky:
* Honor cflags and compiler if set in environment
* Create directories when running make install


plusky:
In building environment people want to specify/extend their own CFLAGS (thats reason why current linuxrc is not build with default distribution flags after all) and it is not unreasonable to give users opportunity to switch different compiler than gcc (i.e gcc-6, clang etc).

plusky:
See also https://build.opensuse.org/request/show/538310 which has to do ugly trick - `CC="cc %{optflags}"` to enforce usage of default distribution flags

wfeldt:
Your patch would **add** to `CFLAGS`. That's rather unusual. I've nothing against introducing some `GCCOPT`, `OPTFLAGS`, or whatever.

plusky:
As for distribution defaults (i,e openSUSE or SLE), CFLAGS is the name environment variable used by default for propagating %{optflags} and reason for this is that usage of CFLAGS is common. Thus if we go for  `CFLAGS`


```
%build
make %{?_smp_mflags}

```
Will work out of box as expected

See also:
1. https://www.gnu.org/software/make/manual/html_node/Environment.html
2. https://www.gnu.org/software/make/manual/html_node/Appending.html
3. https://www.freebsd.org/doc/en/books/porters-handbook/dads-cflags.html

marxin:
I would like to see as well honoring of ```$optflags``` in the package build.

plusky:
cc @kobliha - could you plese check this?

marxin:
I would close this now.

PR #155, opened 2017-11-09
update broken module config even if list is empty

wfeldt:
Otherwise you'll never be able to clear the list entirely.

see https://bugzilla.suse.com/show_bug.cgi?id=1058065#c10


shundhammer:
LGTM

PR #159, opened 2017-11-22
try harder to find a suitable network interface (bsc #1068336)

wfeldt:
After checking network cards, go through the remaining network interfaces.

`sort_a_bit()` ensures we try real hardware first and don't try anything twice.

### DIFF
--- a/url.c
+++ b/url.c
@@ -65,6 +65,9 @@ static slist_t *url_instsys_lookup(char *key, slist_t **sl_ll);
 static char *url_instsys_config(char *path);
 static char *url_config_get_path(char *entry);
 static slist_t *url_config_get_file_list(char *entry);
+static hd_t *find_parent_in_list(hd_t *hd_list, hd_t *hd);
+static hd_t *relink_array(hd_t *hd_array[]);
+static void log_hd_list(char *label, hd_t *hd);
 static hd_t *sort_a_bit(hd_t *hd_list);
 static int link_detected(hd_t *hd);
 static char *url_print_zypp(url_t *url);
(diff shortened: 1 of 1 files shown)


url.c
```diff
@@ -2949,31 +3005,59 @@ hd_t *sort_a_bit(hd_t *hd_list)
...
+    /* 3. wlan cards last */
 
     for(u = 0, hd = hd_list; hd; hd = hd->next) {
       if(!hd->is.wlan) hd_array[u++] = hd;
     }
     for(hd = hd_list; hd; hd = hd->next) {
       if(hd->is.wlan) hd_array[u++] = hd;
     }
-    hd_array[hds] = NULL;
-    for(u = 0; u < hds; u++) hd_array[u]->next = hd_array[u + 1];
```
jreidinger: :+1: for abstracting this relink

url.c
```diff
@@ -1455,23 +1458,23 @@ int url_mount(url_t *url, char *dir, int (*test_func)(url_t *))
...
         break;
 
       case inst_floppy:
-        hw_item = hw_floppy;
+        hw_items = (hd_hw_item_t[]) { hw_floppy, 0 };
         break;
 
       default:
-        hw_item = hw_block;
+        hw_items = (hd_hw_item_t[]) { hw_block, 0 };
```
jreidinger: NP: I see documentation that array have to be NULL ended, so maybe it makes sense to use constant here like `{ hw_block, NULL}` ?

wfeldt: No, that's two different things. Here it's  just an array of enums, what you probably mean is the array of (hd_t *) (the detected  hardware structs).

jreidinger: ah, now I see it. It is structure for hd_list2 method. Then fine for me.

PR #160, opened 2017-12-04
read digests also from CHECKSUMS file

wfeldt:
The new media layout has a 'CHECKSUMS' file instead of the old 'content'.

### DIFF
--- a/file.c
+++ b/file.c
@@ -6,6 +6,8 @@
  *
  */
 
+#define _GNU_SOURCE		/* getline */
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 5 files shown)


url.c
```diff
@@ -2178,9 +2178,10 @@ static int test_is_repo(url_t *url)
   if(!config.keepinstsysconfig) {
     config.digests.failed = 0;
 
-    // check for '/content' or '/repodata/repomd.xml' as indication we have a suse repo
-    // 'content' must be validly signed (we parse it), 'repomd.xml' not (we just check its presence)
-    // zenworks has a different approach ('settings.txt') - they don't have a repo
+    // Check for '/content' resp. '/repodata/repomd.xml' as indication we
+    // have a SUSE repo.
+    // The file must be validly signed (because we parse it).
+    // zenworks has a different approach ('settings.txt') - they don't have a repo.
```
lslezak: NP: I wonder whether we still need the zenworks support...

file.c
```diff
@@ -2705,3 +2707,35 @@ void file_parse_repomd(char *file)
...
+  FILE *fh;
+  char *buf = NULL;
+  size_t buf_size = 0;
+  char sha256[65], name[256];
+  slist_t *sl_digest;
+
+  if(!(fh = fopen(file, "r"))) return;
+
+  while(getline(&buf, &buf_size, fh) > 0) {
+    if(sscanf(buf, "%64s %255s", sha256, name) == 2) {
```
lslezak: NP: maybe we could report an error/warning in the case when a line does not match...

PR #162, opened 2018-01-10
allow kexec to work with secure boot (bsc #1075051)

wfeldt:
For this, the -s option should be used. See discussion in [bsc \#1075051](https://bugzilla.suse.com/show_bug.cgi?id=1075051) for details.


aschnell:
LGTM

PR #163, opened 2018-01-11
extend 'kexec' option to allow updating kernel/initrd only if necessary (bsc #990374)

wfeldt:
This patch adds
- kexec=2: check if instsys does not match, then present a user dialog
- kexec=3: check, but skip the dialog

The default is now kexec=2.


shundhammer:
LGTM

PR #165, opened 2018-03-14
be carefull with 'kexec -s', it's not always available (bsc #1076839)

wfeldt:
Until upstream comes around we need a working solution. Until then, pass
different parameters to kexec.


shundhammer:
LGTM

PR #169, opened 2018-05-30
write s390 config values into /boot/zipl/active_devices.txt (bsc #1095062)

wfeldt:


util.c
```diff
@@ -5555,3 +5555,25 @@ void util_setup_coredumps()
...
 }
 
+
+/*
+ * Write s390 config values into /boot/zipl/active_devices.txt
+ * (see bsc#1095062).
+ */
+void util_write_active_devices(char *format, ...)
+{
+  const char *file_name = "/boot/zipl/active_devices.txt";
```
mvidner: I know that this is the path quoted in the bug report, but shouldn't it be  `/mnt/boot/...` instead?

wfeldt: No. it's in the inst-sys.

lslezak: AFAIK not, the file is copied to `/mnt` later by YaST...

jreidinger: no, it is in insts-sys and we copy it from insts-sys. /mnt should not be ready in this time, not?

lslezak: See https://github.com/yast/yast-installation/pull/705/files

PR #170, opened 2018-06-07
Eliminate chance of read() hang on size check

jjohnson42:
If fd refers to something that another process may be
reading, between select() and read(), something else may consume
the data and result is a stuck read().  Address by making the fd
non blocking and going back to select() when EAGAIN/EWOULDBLOCK
occurs in the read attempt.

### DIFF
--- a/keyboard.c
+++ b/keyboard.c
@@ -482,6 +482,7 @@ void get_screen_size(int fd)
   struct timeval to;
   fd_set set;
   unsigned u1, u2;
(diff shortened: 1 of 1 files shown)


wfeldt:
Thanks!

wfeldt:
I spoke too soon: please redo against the correct branch `master`.

wfeldt:
Forget it, looks like github is just being really slow today.

PR #174, opened 2018-06-21
Eliminate chance of read() hang on size check

wfeldt:
If fd refers to something that another process may be
reading, between select() and read(), something else may consume
the data and result is a stuck read().  Address by making the fd
non blocking and going back to select() when EAGAIN/EWOULDBLOCK
occurs in the read attempt.

This is just https://github.com/openSUSE/linuxrc/pull/170 for sle12.


shundhammer:
LGTM

PR #175, opened 2018-06-21
fix segfault when using an invalid url scheme (bsc #1061824, bsc #1098443)

wfeldt:
Taken from https://github.com/openSUSE/linuxrc/pull/149.


shundhammer:
LGTM

PR #176, opened 2018-06-28
rewrite digest_media_verify() to use new mediacheck lib (bsc #1000947)

wfeldt:
This is basically replacing / removing old code to use `libmediacheck`.

For a `libmediacheck` reference look [here](https://github.com/openSUSE/checkmedia/blob/master/mediacheck.md).


checkmedia.c
```diff
@@ -1,527 +1,182 @@
...
+      (media->iso_blocks & 1) ? ".5" : ""
+    );
   }
 
-  if(iso.digest.got_current) {
-    log_info("  %s: ", iso.digest.name);
-    for(i = 0; i < iso.digest.size; i++) log_info("%02x", iso.digest.full[i]);
-    log_info("\n");
-  }
+  if(media->pad_blocks) log_info("pad: %u kB\n", media->pad_blocks >> 1);
```
jreidinger: above you add .5 here not. Is it imporant info for padding? BTW as majority of sizes are in 512 bytes, maybe to return size in Kb here can be method? like `if(media->pad_blocks) log_info("pad: %u kB\n", size_in_kb(media->pad_blocks));`

checkmedia.c
```diff
@@ -1,527 +1,182 @@
...
-    sprintf(buf + strlen(buf), "\nThis %s is broken.", *iso.media_type == 'C' ? "CD-ROM" : iso.media_type);
-    dia_message(buf, MSGTYPE_ERROR);
-    config.manual=1;
+  if(media->part_blocks) {
+    log_info(
+      "partition: start %u%s kB, size %u%s kB\n",
+      media->part_start >> 1,
+      (media->part_start & 1) ? ".5" : "",
+      media->part_blocks >> 1,
+      (media->part_blocks & 1) ? ".5" : ""
```
jreidinger: as written above I see this multiple time, so it makes sense to create method that return size so it can look like
```
 log_info(
     "partition: start %s kB, size %s kB\n",
     size_in_kb(media->part_start), size_in_kb(media->part_blocks)
)
```

checkmedia.c
```diff
@@ -1,527 +1,182 @@
...
-      if(u == iso.digest.size) iso.digest.got_old = 1;
+  else {
+    if(media->err_block) {
+      sprintf(buf, "Error reading block %u.", media->err_block);
     }
+    else {
+      sprintf(buf, "Checksum wrong.");
+    }
+    sprintf(buf + strlen(buf), "\nThis medium is broken.");
+    dia_message(buf, MSGTYPE_ERROR);
```
jreidinger: buf is used only in this context. Is possible to move definition here? As it is fixed size buffer, I feel safer if it is closer to its usage so harder to overlook if more or longer messages added to it.

checkmedia.c
wfeldt: padding is always in 2k units, so no odd block numbers possible

checkmedia.c
wfeldt: Not so easy in C; I can't use a static buffer and with dynamic mem allocation complexity blows up.

I'd leave it as is (it's basically the same code as in the checkmedia program).

PR #182, opened 2018-11-27
Added new BuildRequires to devel doc

shundhammer:
linuxrc doesn't build without that package. It's there in the .spec file, but not in the instructions in the README.md.


wfeldt:
lgtm

PR #183, opened 2018-11-28
Added parameter norepo (fate#325482)

shundhammer:
Documentation:

https://en.opensuse.org/SDB:Linuxrc

> NoRepo
>
> Disable the repo location check and do not write a ZyppRepoURL to /etc/install.inf; expect YaST to take care of the repo selection.
>
> This is useful for delegating the repo selection to the registration server so a user can simply enter a registration code first, then the matching product will be automatically selected based on that registration code, and the respective repos will be addded accordingly (Fate#325482).


wfeldt:
lgtm

PR #186, opened 2019-03-18
Fix usage of a out-of-scope vector constructor.

marxin:
GCC 9 starts to optimize undefined behaviour code as seen here:
https://www.gnu.org/software/gcc/gcc-9/porting_to.html#complit


wfeldt:
lgtm

PR #192, opened 2019-07-17
Support for a multi-repository installation medium (bsc#1141889)

lslezak:
- Added support for a multi-repository installation medium (like the Packages DVD)
- Do not fail when the installation medium does not contain a SUSE tags or a RPM-MD repository but contains the repositories in subdirectories (indexed in the `/media.1/products` file)
- https://bugzilla.suse.com/show_bug.cgi?id=1141889
- Tested manually with a custom built DVD installation image (Packages DVD + copied boot files from the Installer DVD + updated linuxrc)
  - Works fine, does not complain about a missing repository
  - The `/etc/install.inf` contains the correct `ZyppRepoURL` value as expected
- Updated the `Dockerfile` to use the newer image built directly in OBS
- Updated documentation
  - SUSE tags repositories are obsolete since SLE15/Leap15
  - Documented multi-repository medium
  - Documented `no_repo=1` behavior

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM yastdevel/cpp
+FROM registry.opensuse.org/yast/head/containers/yast-cpp:latest
 
 RUN zypper --non-interactive in --no-recommends \
   e2fsprogs-devel \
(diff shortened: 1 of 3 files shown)


wfeldt:
lgtm

wfeldt:
lgtm

PR #194, opened 2019-08-14
choose correct kernel image on aarch64 (bsc#1145625)

wfeldt:
## Problem

https://bugzilla.suse.com/show_bug.cgi?id=1145625

'Boot linux system' boot option does not work on aarch64.

## Analysis

When booting 'installed system' linuxrc has to pick the correct kernel/initrd pair. Unfortunately the kernel naming scheme differs for each architecture.

## See also

[Original fix in `master`](https://github.com/openSUSE/linuxrc/pull/193)


wfeldt:
lgtm

PR #197, opened 2019-11-15
Add linuxrc reboot_timeout option

teclator:
@okurz is adding some patches in order to set the reboot timeout through linuxrc

See https://bugzilla.suse.com/show_bug.cgi?id=1122493

This PR is just for helping him with the changes needed.

See: https://github.com/yast/yast-yast2/pull/977 & https://github.com/yast/yast-installation/pull/823


wfeldt:
I would not do it this way. To pass options to yast linuxrc doesn't have to know about, simply use the `ptoptions` option. See https://en.opensuse.org/SDB:Linuxrc#p_ptoptions. Just add `reboot_timeout` here: https://github.com/openSUSE/installation-images/blob/master/data/initrd/theme.file_list#L78.

teclator:
Replaced by https://github.com/openSUSE/installation-images/pull/344

PR #203, opened 2019-12-04
Added logic to linuxrc.c and util.c to retrieve the name of the proce…

markkp:
…ssorlinuxrc is running on, for mainframe systems only.

This is for feature jsc#SLE-9474. I have compiled it on all supported architectures, but only tested it on s390x. There will also need to be a change to the installation images to add /usr/lib64/libqc.so.2.0.0, and the symlink of /usr/lib64/libqc.so.2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 CC	= gcc
 CFLAGS	= -c -g -O2 -Wall -Wno-pointer-sign
+ARCH	= $(shell /usr/bin/uname -m)
+ifeq ($(ARCH),s390x)
+LDFLAGS	= -rdynamic -lhd -lblkid -lcurl -lreadline -lmediacheck -lqc
+else
 LDFLAGS	= -rdynamic -lhd -lblkid -lcurl -lreadline -lmediacheck
+endif
 
 GIT2LOG := $(shell if [ -x ./git2log ] ; then echo ./git2log --update ; else echo true ; fi)
 GITDEPS := $(shell [ -d .git ] && echo .git/HEAD .git/refs/heads .git/refs/tags)
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,9 @@
 CC	= gcc
 CFLAGS	= -c -g -O2 -Wall -Wno-pointer-sign
 ARCH	= $(shell /usr/bin/uname -m)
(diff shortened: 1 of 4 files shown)


markkp:
Hmm. If I'm not mistaken, the initrd already has /usr/lib64/libqc.so.2* in it. Could someone verify that, so I don't spend time trying to work up a pull request for that, also? Thanks.

Makefile
```diff
@@ -1,6 +1,11 @@
 CC	= gcc
 CFLAGS	= -c -g -O2 -Wall -Wno-pointer-sign
+ARCH	= $(shell /usr/bin/uname -m)
+ifeq ($(ARCH),s390x)
+LDFLAGS	= -rdynamic -lhd -lblkid -lcurl -lreadline -lmediacheck -lqc
+else
 LDFLAGS	= -rdynamic -lhd -lblkid -lcurl -lreadline -lmediacheck
+endif
```
wfeldt: Please do something like
```
ifeq ($(ARCH),s390x)
LDFLAGS += -lqc
endif
```
so the arch independent flags can stay in a single separate line.

linuxrc.c
```diff
@@ -895,10 +917,18 @@ void lxrc_init()
   if(!config.had_segv) {
     if (config.linemode)
       putchar('\n');
+#ifdef __s390x__
+    printf(
+      "\n>>> %s installation program v" LXRC_FULL_VERSION " (c) 1996-2019 SUSE LLC on %s <<<\n",
+      config.product,
+      config.hwp.machine_name
+    );
+#else
```
wfeldt: That's a lot if `ifdef __s390x__` and copy-and-paste code duplication.

I would suggest something like having a new var like `config.machine` on *all* archs that is initialized with `""` and on s390x gets `" on Foo Bar"`. The string can then be inserted unconditionally in all these `printf`s where it matters.

This also makes it easy later to have something similar on non-s390.

wfeldt:
And please insert the string `jsc#SLE-9474` somewhere in the first line of the git commit message.

The package changelog is auto-generated from that line and the Jira reference is needed for OBS tooling.

markkp:
OK, I'll submit again.

PR #204, opened 2019-12-05
Append EXTRA_CFLAGS to CFLAGS.

marxin:
I would like the package would respect $Optflags in OBS.
Thus I'm suggesting to append EXTRA_CFLAGS to CFLAGS. One can
then 'export EXTRA_CFLAGS=$Optflags'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 CC	= gcc
-CFLAGS	= -c -g -O2 -Wall -Wno-pointer-sign
+CFLAGS	= -c -g -O2 -Wall -Wno-pointer-sign $(RPM_OPT_FLAGS)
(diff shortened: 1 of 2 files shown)


marxin:
Which is probably what was already suggested in https://github.com/openSUSE/linuxrc/pull/154.

wfeldt:
Doesn't simply adding `$(RPM_OPT_FLAGS)` do what you want?

marxin:
> Doesn't simply adding `$(RPM_OPT_FLAGS)` do what you want?

That would work. So should I add the `$(RPM_OPT_FLAGS)` to `CFLAGS` in the `Makefile`?

wfeldt:
if you change this pr, that would be ok

marxin:
> if you change this pr, that would be ok

Good, done.

PR #207, opened 2019-12-13
Revert "Append RPM_OPT_FLAGS to CFLAGS." (bsc#1158996)

wfeldt:
## Problem

https://bugzilla.suse.com/show_bug.cgi?id=1158996

On some ppc64le setups, linuxrc runs into "stack smashing detected" and crashes.

## Solution

Revert https://github.com/openSUSE/linuxrc/pull/204. This avoids compiling with `$RPM_OPT_FLAGS`.

That's of course not the final solution but it gets linuxrc going until the real cause has been identified.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 CC	= gcc
-CFLAGS	= -c -g -O2 -Wall -Wno-pointer-sign $(RPM_OPT_FLAGS)
+CFLAGS	= -c -g -O2 -Wall -Wno-pointer-sign
 LDFLAGS	= -rdynamic -lhd -lblkid -lcurl -lreadline -lmediacheck
 ARCH	= $(shell /usr/bin/uname -m)
 ifeq ($(ARCH),s390x)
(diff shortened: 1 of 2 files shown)


foursixnine:
I'm late, but LGTM too :D https://openqa.suse.de/tests/3704503#step/user_settings/1

scarabeusiv:
are you guys crazy? the srack smashing needs to be fixed not hidden by removing our hardening flags.

mvidner:
@scarabeusiv You are right, but we needed a workaround to unblock testing for a class of machines. We're scheduling a proper fix right now.

mvidner:
The followup is https://trello.com/c/UpuZNwxv

plusky:
The followup gives me "Card not found."

jsrain:
The Trello is team-internal. As the real bug needs more work to find out the real cause of the problem, I reopened it for future tracking.

wfeldt:
For the record, the final solution: https://github.com/openSUSE/linuxrc/pull/209

PR #210, opened 2020-01-28
For bsc#1161701, Moved the logic to retrieve the platform name into a

markkp:
subroutine, and moved the invocation of it until after the call to
lxrc_add_parts so that the squashfs files have been loopback mounted.

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -1759,12 +1759,12 @@ void select_repo_url(char *msg, char **repo)
 
 char * get_platform_name()
 {
+  char *platform = NULL;
 #if defined(__s390__) || defined(__s390x__)
   void *qc_configuration_handle = NULL;
   const char *qc_result_string = NULL;
   int qc_return_code = 0;
   int qc_get_return_code = 0;
-  char *platform = NULL;
 
   qc_configuration_handle = qc_open(&qc_return_code);
   if (qc_return_code==0)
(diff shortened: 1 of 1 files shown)


linuxrc.c
```diff
@@ -1775,3 +1757,30 @@ void select_repo_url(char *msg, char **repo)
...
 }
 
+char * get_platform_name()
+{
+#if defined(__s390__) || defined(__s390x__)
+  void *qc_configuration_handle = NULL;
+  const char *qc_result_string = NULL;
+  int qc_return_code = 0;
+  int qc_get_return_code = 0;
+  char *platform = NULL;
```
wfeldt: that should go outside the `#if`

linuxrc.c
```diff
@@ -1775,3 +1757,30 @@ void select_repo_url(char *msg, char **repo)
...
+
+  if (qc_get_return_code<=0) {
+    log_show("Unable to retrieve machine type.\n");
+    strprintf(&platform, "%s", "on unknown machine type");
+  }
+  else strprintf(&platform, "on %s", qc_result_string);
+
+  qc_close(qc_configuration_handle);
+#else
+  &platform="";
```
wfeldt: better `str_copy(&platform, "")` as not to mix allocated and static strings

wfeldt:
And, while you're at it, could you upgrade the year in that `... 1996-2019 SUSE LLC ...` startup message to 2020?

markkp:
Hopefully these changes are what you meant. Please let me know if I'm wrong.

PR #211, opened 2020-02-04
(bsc#1161701) Moved all the logic for displaying the processor

markkp:
model name from linuxrc.c to install.c so that the symlinks from /usr/lib64/gconv that point to
/mounts/mp_0000/usr/lib64/gconv are in place before making the call to get_platform_name.

Removed config.platform_name printing from util.c

### DIFF
--- a/install.c
+++ b/install.c
@@ -46,6 +46,9 @@
 #include "auto2.h"
 #include "url.h"
 #include "checkmedia.h"
+#ifdef __s390x__
+#include <query_capacity.h>
+#endif
 
 #ifndef MNT_DETACH
 #define MNT_DETACH	(1 << 1)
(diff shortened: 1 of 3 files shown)


wfeldt:
No, that's the wrong way. The original patch is ok.

The locale and gconv files are not part of the initrd because they were not needed so far. If they are really needed they have to be moved there from the inst-sys image [here](https://github.com/openSUSE/installation-images).

But why on earth would a tool need charset conversion to provide the machine name???

markkp:
While Linux runs fully ASCII on the mainframe, the hardware itself does not. So, things that need to be read from the hardware to be displayed on Linux have to be converted from EBCDIC to ASCII.

PR #212, opened 2020-02-07
fix rescue system setup (bsc#1160378)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1160378

systemd does not work properly because there's no mountpoint for `/`

## Solution

- fix rescue system setup to ensure it has a mount point for `/`

## Bonus

- removes some obsolete code

## Todo

- for yet unknown reasons systemd removes `/lib/modules` and `/lib/firmware` and then complains that it is not able to load kernel modules

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -295,8 +295,8 @@ void lxrc_change_root()
   slist_t *sl;
   char *argv[3] = { };
   char *dirs[] = {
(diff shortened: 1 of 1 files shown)


wfeldt:
closed in favor of https://github.com/openSUSE/linuxrc/pull/216

PR #217, opened 2020-03-16
Try suboption for ifcfg

mchf:
https://trello.com/c/WJmotQTM/1548-5-feature-sle-8965-additional-try-option-for-linuxrc-ifcfg-network-setup

## Problem ##

We were asked for implementing a feature, when static configuration is tried on net devices one-by-one until net installation source can be reached.

## Solution ##

The suboption makes sense only for static configuration with pattern instead of particular device. Each ifcfg configuration is assigned to each matching device. After assigning another device, we restart installation  and check whether it can reach the installation image. Only ifcfg configs marked with try suboption are handled when searching matching device.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ before_install:
   - docker run --rm -it linuxrc-image rpm -qa | sort
 
 script:
-  - docker run --rm -it -e TRAVIS=1 linuxrc-image bash -c "make -j `nproc` && ./smoke_test.sh"
+  - docker run --rm -it -e TRAVIS=1 --privileged linuxrc-image bash -c "make -j `nproc` && ./smoke_test.sh"
--- a/auto2.c
+++ b/auto2.c
@@ -136,6 +136,7 @@ int auto2_init()
         else
         {
           log_debug("auto2_init: Cannot find another device to try");
+          break;
         }
       }
     }
--- a/auto2.c
+++ b/auto2.c
@@ -611,7 +611,7 @@ void auto2_user_netconfig()
   if(!net_config_needed(0)) return;
 
   check_ptp(config.ifcfg.manual->device);
-  
+
   if( ((net_config_mask() & 3) == 3) || (config.ifcfg.manual->ptp && ((net_config_mask() & 1) == 1)) ) {
     /* we have IP & netmask (or just IP for PTP devices) */
     config.net.configured = nc_static;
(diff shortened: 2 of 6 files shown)


wfeldt:
Will check

wfeldt:
If I'm not mistaken, the idea is to run `auto2_init()` for each attempt and see if that works.

That won't work for several reasons - basically the expectation everywhere is that
`auto2_init()` is run only once (hence the 'init'):

- the s390 setup is included there - you want that only once
- the mediacheck call is in there - also only once
- in manual mode (`config.manual == 1`) it must **not** be called - manual is the fallback in case auto does not work
- `auto2_init()` may fail for network-unrelated reasons (like broken local install repo)

You might be tempted to go for `auto2_find_repo()` instead, then. But I think this will also not do: you have to take care that `dud`, `sshd`, and `info` with network URL also trigger the search process.

All in all, I still believe my original proposal (provided it actually works) is the most rational approach here.

wfeldt:
Generally, looks ok; but let me run some tests first.

wfeldt:
It does not work. Or let's say it does not do what one would expect.

First, it's not easy to setup a vm for this. linuxrc will try connected interfaces first - so it's not easy to get a scenario where you would try some failing interfaces and then only later hit the right one.

Here's what happens when there are 3 disconnected interfaces (and no connected one).
Commandline is
```
install=ftp://10.10.42.22/cd1 ifcfg=*=try,10.0.2.15/24,10.0.2.2
```
linuxrc tries 5 network configs in turn:
- ens4-static
- ens3-dhcp
- ens5-dhcp
- ens3-static
- ens5-static

What I would expect - and that's exactly what happens when you use the equivalent
```
install=ftp://10.10.42.22/cd1 hostip=10.0.2.15/24 gateway=10.0.2.2
```
- ens3-static
- ens4-static
- ens5-static

So I'm back to my original argument that the very simple and effective way to achive the goal might be to copy the ifcfg with `try` set into `config.ifcfg.manual`.

wfeldt:
I'll look at it.

wfeldt:
I've taken the parser and just copied the config. This didn't fully work as expected and uncovered a bug in the device iteration loop. After fixing that, this seems to do:

https://github.com/openSUSE/linuxrc/pull/219

mchf:
Well, I have to admit, that this solution behaves this way bcs it involves also original fallback which tries to configure device(s) with dhcp, and new static configuration is tried later. I've talked about it in the irc .... But lets move to the other solution.

PR #218, opened 2020-03-30
Update doc for hostname in different SLE versions

SergioAtSUSE:


linuxrc_hostname.md
```diff
@@ -1,30 +1,35 @@
...
+- Before SLE12, of the form `linux-XXXX`
+- Since SLE12, `linux-XXXX.suse`
+- Since SLE15-SP2, `linux`
+`XXXX` is a random-generated alphanumerical string.
 
-When linuxrc doesn't set `/etc/install.inf::Hostname` yast generates a
-default `/etc/hostname` entry of the form `linux-XXXX` (+ `.suse` in sle12).
-`XXXX` is some random part.
+linuxrc doesn't generate the `/etc/hostname` file in the installation
+environment. Yast generates it.
```
wfeldt, line 27: Really? I don't see one.

SergioAtSUSE: > Really? I don't see one.

Ok, it sounds like Yast always created. I wanted to express that if it is present, it is created by Yast.

That happens when linuxrc::hostname is used, right?

linuxrc_hostname.md
```diff
@@ -1,30 +1,35 @@
...
-`/etc/install.inf::Hostname` and ends up in `/etc/hostname` of the installed system.
+When yast is started, the value from (1) is passed to yast via
+`/etc/install.inf::Hostname` and ends up in `/etc/hostname` of the installed
+system.
+
+When linuxrc doesn't set `/etc/install.inf::Hostname`, yast generates a default
+`/etc/hostname` entry:
+- Before SLE12, of the form `linux-XXXX`
+- Since SLE12, `linux-XXXX.suse`
+- Since SLE15-SP2, `linux`
```
teclator, line 23: YaST should not propose a hostname if no one was configured explicitly according to documentation

linuxrc_hostname.md
```diff
@@ -1,30 +1,35 @@
...
+`XXXX` is a random-generated alphanumerical string.
 
-When linuxrc doesn't set `/etc/install.inf::Hostname` yast generates a
-default `/etc/hostname` entry of the form `linux-XXXX` (+ `.suse` in sle12).
-`XXXX` is some random part.
+linuxrc doesn't generate the `/etc/hostname` file in the installation
```
teclator, line 26: Well, it is not a generation at all. When a `hostname` is configured through linuxrc, yast writes the given hostname to /etc/hostname at very beginning of the installation.

https://github.com/yast/skelcd-control-leanos/blob/master/control/control.leanos.xml#L516
https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/inst_install_inf.rb#L19
https://github.com/yast/yast-network/blob/c1ed7b6ca13a391b35a7de95458bcc628da26084/src/lib/network/install_inf_convertor.rb#L61

linuxrc_hostname.md
SergioAtSUSE: But, it does, because `SetHostname=1` by default. I expressed it wrong.
WDYT about the new wording?

linuxrc_hostname.md
wfeldt: > Ok, it sounds like Yast always creates it

No! Nobody creates it. The document said

> There is no `/etc/hostname` file in the installation environment.

And that is correct (it's an empty file, to precise).

SergioAtSUSE: /etc/hostname is a filename, so it is a file
empty file and no file are different things.

wfeldt: An empty /etc/hostname and a missing one are semantically equivalent, afaik.

But my point was that yast does not create it. The emtpy file is always there and comes directly from the netcfg package.

If you want to push your point and stress the difference between empty and missing /etc/hostname feel free to do so, it's developer documentation after all.

But I don't want the doc to end up less accurate than before.

linuxrc_hostname.md
wfeldt: > YaST writes the given hostname to /etc/hostname at very beginning of the installation

I don't see that happen in my test. /etc/hostname **in the installation system** is always empty.

teclator: At which point are you checking it?

teclator: In which release?, and have you checked once YaST starts?

teclator: As described, it is set by the very first client of the installation workflow.

Screenshot_sles12sp3_2020-04-06_09:08:39
Screenshot_sles12sp3_2020-04-06_09:10:04

teclator: And the same in SLE-15:

Screenshot_sle15_2020-04-06_09:48:29

wfeldt: Strange, but you are correct. No idea what I was looking at.

linuxrc_hostname.md
```diff
@@ -1,31 +1,76 @@
 # hostname setting in installation/rescue system
 
-There are 3 cases:
+The hostname can be set (in order of precedence):
 
-1. no hostname is set
+1. via `hostname` boot option
+2. via DHCP (either from HOSTNAME or FQDN)
+3. Default hostname set by Yast
```
teclator, line 7: We should take into account that this is a linuxrc documentation, and thus, the previous documentation was still valid.

From my POV, main changes should address the write of /etc/hostname by YaST during a common installation as well as the proposal of the hostname by YaST at the end of the installation previously described.

What I mean is that linuxrc still sets the transient hostname ('install' or the one given by hostname='hostname').

linuxrc_hostname.md
```diff
@@ -1,31 +1,76 @@
...
-linuxrc defaults to `install`, which may be overridden by dhcp (2) or user
-(3). (3) takes precedence over (2).
+- `etc/hostname` is an empty file at the beginning of the installation.
+- `linuxrc` doesn't write anything to `/etc/hostname`.
+- `linuxrc` writes by default `/etc/install.inf::SetHostname=1`.
 
-## installation system
+## if `hostname` boot option is used
+
+1. linuxrc takes the value of `hostname` boot option and writes it to
```
teclator, line 21: It also sets the current hostname with the one given

linuxrc_hostname.md
```diff
@@ -1,31 +1,76 @@
...
+   `/etc/install.inf::Hostname`.
+2. linuxrc writes `/etc/install.inf::SetHostnameUsed=1`
+
+#### Example
+
+`hostname=myhost` is used
+
```
teclator, line 28: `hostname myhost`

linuxrc_hostname.md
```diff
@@ -1,31 +1,76 @@
...
+```
+[...]
+SetHostname: 1
+SetHostnameUsed: 1
+Hostname: myhost
+[...]
+```
+
+## If `hostname` boot option is not used
+
```
teclator, line 39: `hostname install`

linuxrc_hostname.md
```diff
@@ -1,31 +1,76 @@
...
+**/etc/install.inf**
+```
+[...]
+SetHostname: 1
+SetHostnameUsed: 0
+[...]
+```
+
+## If the DHCP server is offering a hostname
+
```
teclator, line 56: if linuxrc configures DHCP then it could add it explicitly to the configured interface.

But a YaST client will check also the default from the control file and will adjust the sysconfig configuration:

see https://github.com/yast/yast-network/pull/564

linuxrc_hostname.md
SergioAtSUSE: ## Contradiction?

I see a contradiction between the first to paragraphs.

> ...this is a linuxrc documentation...

vs 
> ...main changes should address the write of /etc/hostname by YaST...

and
> ...the proposal of the hostname by YaST at the end of the installation...


## Suggestion

I think that it would be better to either

1. split the documentation between everything linuxrc related in openSUSE/linuxrc/linuxrc_hostname.md and everything Yast related in Yast/yast2-network? Yast/yast2-installer? and crosslink them,
2. or put everything in openSUSE/linuxrc/linuxrc_hostname.md. If not the doc will be technically true but confusing/misleading and useless.

> ...linuxrc still sets the transient hostname...
ah! I thanks for clarifying that :)

> ...[linuxrc] also enables the DHCLIENT_SET_HOSTNAME in case of configured dhcp interface...

Right, I forgot about that.

linuxrc_hostname.md
SergioAtSUSE: I don't understand. I was trying to  show how a user would write the parameter as boot parameter.
What does it mean `hostname myhost`?

linuxrc_hostname.md
SergioAtSUSE: So, linuxrc always sets the transient hostname before Yast in the installation environment?

linuxrc_hostname.md
SergioAtSUSE: Is current hostname a synonym of transient hostname?

linuxrc_hostname.md
wfeldt: > In general, I would split the documentation, leaving here only linuxrc responsibilities but not mixing with YaST workflows.

The document title is 'hostname setting in installation/rescue system' and I would like to stick to that here. As you see it's complex enough. :-)

linuxrc_hostname.md
SergioAtSUSE: So, the name given by DHCP server will set as transient hostname on the installation environment by Yast, right?

linuxrc_hostname.md
mchf: empty file is incorrect and startup scripts handle it as an exception - "do not set hostname". If you try to load empty /etc/hostname in running system (e.g. using hostname -F) you'll find that it behaves differently - reports invalid hostname (for SLE-15*, for having more fun in SLE-12* it works differently - it unsets the hostname there).

linuxrc_hostname.md
mchf: well ... yes and no. I'll try to simplify it ... Basically you have "static hostname" - the hostname which is set according to /etc/hostname and "transient hostname" - hostname which is e.g. set from dhcp or e.g. using ```hostname <hostname>```. 

What is important here is that linuxrc sets transient hostname which means that the hostname is not written into ```/etc/hostname/``` and will not survive reboot.

Sorry for being pedantic but we have had lot of issues with that during past years bcs of lot of misunderstanding.

linuxrc_hostname.md
mchf: yes.

linuxrc_hostname.md
SergioAtSUSE: yes, but, what do you refer to as "current hostname"?

SergioAtSUSE:
> Content is ok, but could you unify at least the 'if' spelling in the section titles ('if' vs. 'If')?

Done

wfeldt:
Thanks for your patience!

PR #232, opened 2020-10-28
fix linuxrc module option parsing

wfeldt:
## Problem

If linuxrc sees options of the form `foo.bar=xxx` it assumes `bar=xxx` to be a kernel module option for `foo`.

This was broken as linuxrc did this sometimes also when `foo` was a valid linuxrc option.


shundhammer (approved):
I am clearly not an expert on this, but LGTM.

PR #237, opened 2021-01-05
Since kernel 5.10, /sys/firmware/efi/vars has been dropped

ggardet:
for non-x86 architectures, only /sys/firmware/efi/efivars remains
So, check both to know if we are running on EFI boo#1180408

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -1078,7 +1078,7 @@ void lxrc_init()
   if(config.had_segv) config.manual = 1;
 
   /* check efi status */
-  if(util_check_exist("/sys/firmware/efi/vars") == 'd') {
+  if(util_check_exist("/sys/firmware/efi/vars") == 'd' || util_check_exist("/sys/firmware/efi/efivars") == 'd') {
     config.efi_vars = 1;
   }
   log_debug("efi = %d\n", config.efi_vars);


Vogtinator:
With older uboot, there's no `efivars` either, so I imagine that there are cases where neither of those is present. What about checking for `/sys/firmware/efi/`?

ggardet:
> With older uboot, there's no `efivars` either, so I imagine that there are cases where neither of those is present. What about checking for `/sys/firmware/efi/`?

Would it be enough?

Vogtinator:
> > With older uboot, there's no `efivars` either, so I imagine that there are cases where neither of those is present. What about checking for `/sys/firmware/efi/`?
> 
> Would it be enough?

It seems like YaST can handle EFI systems without access to efivars, so I think so. It shouldn't be present on systems which weren't EFI booted.

dirkmueller:
> It seems like YaST can handle EFI systems without access to efivars, so I think so. It shouldn't be present on systems which weren't EFI booted.

Looking at the grub2 code and other places, the code seems to be checking if the directory is nonempty (not if it exists). as such that sounds unsafe (it sounds like there are conditions where the directory exists but the efivars filesystem isn't mounted). see

https://github.com/openSUSE/grub2/blob/2df291226638261d50fadcab1f5edb6c12ab6cfd/grub-core/osdep/linux/platform.c#L114-L115

if we want to be nitpicking, we could be changing the check away from /vars to just validate /efivars as that is where the efivarsfs is mounted to.

jreidinger:
@dirkmueller it is handled in yast2-bootloader which pass additional parameters if EFI is wanted and efivars is missing. https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/grub_install.rb#L104

wfeldt:
Given all that back and forth in [bsc#1174029](https://bugzilla.suse.com/show_bug.cgi?id=1174029) with no `efivars`, empty `efivars`, read-only `efivars`, 'normal' `efivars` I feel the patch is as good as it gets.

AIUI aarch64 is always using EFI anyway...

wfeldt (approved):
lgtm

PR #238, opened 2021-01-11
Reduce use of stack based buffers for string operations

dirkmueller:
strprintf() allocates the right memory size for us, so we don't need static buffers and strdup()
    
Also make use of snprintf() to avoid overflows.
    
Last but not least free(NULL) is safe, so remove redundant if() checks.

### DIFF
--- a/module.c
+++ b/module.c
@@ -1,10 +1,9 @@
-#define _GNU_SOURCE
-
 /*
  *
  * module.c      Load modules needed for installation
  *
- * Copyright (c) 1996-2002  Hubert Mantel, SuSE Linux AG  (mantel@suse.de)
+ * Copyright (c) 1996-2002  Hubert Mantel, SuSE Linux AG
+ * Copyright (c) 2002-2021  SUSE Solutions GmbH
  *
  */
 
@@ -126,7 +125,7 @@ int mod_has_module_ext(const char *file, const size_t len, size_t *ext_pos)
   for (mod_ext = mod_extensions; mod_ext->ext; mod_ext++)
     if (mod_cmp_ext(file, len, mod_ext->ext, mod_ext->len)) {
       if (ext_pos)
-	*ext_pos = len - mod_ext->len;
+        *ext_pos = len - mod_ext->len;
       return 1;
     }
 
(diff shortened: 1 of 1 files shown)


wfeldt:
Is that the solution for https://bugzilla.suse.com/show_bug.cgi?id=1180792?

dirkmueller:
@wfeldt see comment in bugzilla. I can't really reproduce the issue in test mode for some reason.. I can't figure out where to get the directory /lib/modules/$kernelv/initrd that it seems to read in the installation sys into my testmode setup. 

so these changes are just from manual code review. there are piles of potentially overflowable buffers, really hard to decide where to start.. :(

wfeldt:
I wasn't sure which was first. :-)

I'll have a look.

wfeldt:
While code might not look ideally, I don't think it's wrong. The reported problem gets fixed with https://github.com/openSUSE/linuxrc/pull/241.

wfeldt:
I spoke too soon. There is a bug. :-(

dirkmueller:
@wfeldt wow, indeed, this looks like an overflow. still I think this code is a worthwhile cleanup (especially the removal of the temporary 15 bytes stack buffer in read_module_list)

wfeldt:
I believe I've now covered everything in https://github.com/openSUSE/linuxrc/pull/241 but for the WiFi issue you reported.

I will commit the bug fixes first and then revisit this pr for the cleanup part. One hunk is already covered in my pr; the rest I'll have to check.

dirkmueller:
@wfeldt so I realized that there is already a function for this, strprintf() from utils, so for consistency I remove the asprintf stuff again.

module.c
```diff
@@ -267,26 +266,19 @@ void mod_update_list()
...
   while((de = readdir(d))) {
-    i = strlen(de->d_name);
-    if(
-      i < (int) sizeof buf &&
-      mod_has_module_ext(de->d_name, i, &ext_pos)
-    ) {
-      strcpy(buf, de->d_name);
-      buf[ext_pos] = 0;
+    if(mod_has_module_ext(de->d_name, strlen(de->d_name), &ext_pos)) {
+      int found = 0;
```
wfeldt, line 277: This is not correct as it edits out that `buf[ext_pos] = 0`. This was actually the reason for that duplicate buffer.

wfeldt:
I've taken your code and extended it to remove all static string buffers and avoid `snprintf` completely: https://github.com/openSUSE/linuxrc/pull/244.

dirkmueller:
thanks!

PR #239, opened 2021-01-13
fix detection of config file type (XML vs. plain text) bsc#1179936

wfeldt:
## Issue

- https://bugzilla.suse.com/show_bug.cgi?id=1179936
- https://trello.com/c/zJMEV1QY

linuxrc doesn't restrict itself to the `<info_file>` section when parsing an AutoYaST file.

## Solution

Improve the detection of the config file type. Make resaonably sure it sticks to the `<info_file>` section if the file is a human-readable formatted XML.

The next escalation step would be to link against libxml but this seems not yet to be necessary.

This also improves logging - at log level 2 the entire parsed config file is logged.

Recognized valid formattings are

```
<info_file>
  foo=10
  bar=200
</info_file>

<info_file></info_file>

<info_file/>
```

Do **not** put the element on a single line **if it has content** (it will be ignored):

```
<info_file>foo=10</info_file>
```

### DIFF
--- a/file.c
+++ b/file.c
@@ -606,29 +606,50 @@ char *file_read_info_file(char *file, file_key_flag_t flags)
  */
 void file_do_info(file_t *f0, file_key_flag_t flags)
 {
-  file_t *f;
+  file_t *f, *xml_section_start = 0;
   int i, is_xml = 0;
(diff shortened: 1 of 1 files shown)


schubi2 (approved):
LGTM

PR #241, opened 2021-01-15
fix writing out of menu item array bounds (bsc#1180792)

wfeldt:
## Problems

- https://bugzilla.suse.com/show_bug.cgi?id=1180792
- https://trello.com/c/x1LXNeEn

1. When leaving the `Expert -> Kernel Modules (Hardware Drivers)` menu a stack smashing might be detected. The internal array was one item too small in some situations.
1. If a module has a very long description (try `modinfo sbc_epx_c3`) a buffer overrun might occur.
1. In some code paths module loading missed the actual module loading.

## Solution

1. Rewriting the code to guard against this in future. It's still a static array but reasonably ensured the length is correct.
1. use `asprintf`
1. fix code flow

### DIFF
--- a/module.c
+++ b/module.c
@@ -1,3 +1,5 @@
+#define _GNU_SOURCE
+
 /*
  *
  * module.c      Load modules needed for installation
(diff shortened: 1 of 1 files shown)


shundhammer (approved):
LGTM

PR #244, opened 2021-01-19
cleanup module code to avoid static string buffers (bsc#1180792)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1180792
- https://trello.com/c/x1LXNeEn

Code cleanup to avoid static string buffers in module code.

This is a follow-up to https://github.com/openSUSE/linuxrc/pull/241 and extends https://github.com/openSUSE/linuxrc/pull/238.

### DIFF
--- a/module.c
+++ b/module.c
@@ -147,7 +147,7 @@ char *mod_basename(const char *file)
   char *buf = strdup(file);
   int base_len = mod_basename_len(file);
 
-  if(base_len) buf[base_len] = 0;
+  if(buf && base_len) buf[base_len] = 0;
 
   return buf;
 }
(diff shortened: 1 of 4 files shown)


module.c
```diff
@@ -119,98 +117,61 @@ static inline int mod_cmp_ext(const char *file, const size_t len,
...
+  char *buf = strdup(file);
+  int base_len = mod_basename_len(file);
 
-  file_len = sprintf(file, "%s/%s", prefix, module);
-  for (mod_ext = mod_extensions; mod_ext->ext; mod_ext++) {
-    strcpy(file + file_len, mod_ext->ext);
-    if(util_check_exist(file))
-      return 1;
-  }
+  if(base_len) buf[base_len] = 0;
```
shundhammer, line 150: If strdup() failed (out of memory), this will segfault.
I suggest to check buf immediately after strdup() and return 0 if buf == 0.

shundhammer (approved):
Just the one pathological case with a possible segfault when that one particular strdup() fails, otherwise LGTM.

wfeldt:
Just a note: `strprintf` is a convenience wrapper around `asprintf` to take care of the undefined state if `asprintf` failes.

shundhammer (approved):
LGTM

PR #245, opened 2021-01-19
Replace splash with plymouth utils (boo#1149070)

hellcp:


wfeldt (approved):
Basically ok but I need a bug or jira reference like bsc#123456 in the title. Else the submission will get blocked later on.

PR #247, opened 2021-03-04
fix buffer overflow when setting up loop device (bsc#1180161)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1180161
- https://trello.com/c/0zqcWVhD

When linuxrc needs to set up a loop device it crashes if the file name is a bit longish (>= 64 chars).

This happens because the kernel's `loop_info` struct has a file name field `char lo_name[LO_NAME_SIZE]` with `LO_NAME_SIZE = 64`.

## Solution

Use `strncpy`.

## Bonus

Check remaining `strcpy` uses.

### DIFF
--- a/auto2.c
+++ b/auto2.c
@@ -274,7 +274,7 @@ void auto2_scan_hardware()
       if(hd->bus.id == bus_usb) ju++;
       di = hd->driver_info;
       if(di && di->any.type == di_kbd) {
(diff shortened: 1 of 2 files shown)


markkp (approved):

wfeldt:
Thanks for checking!

PR #250, opened 2021-03-09
add 'extend' option (bsc#1183368)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1183368
- https://github.com/yast/yast-installation/issues/925

Add `extend` option to linuxrc to be able to load additional inst-sys parts. An inst-sys part is any file that is available on the installation medium in the `/boot/ARCH` directory.

The argument is a comma (',') separated list. If a part is preceded with a '?' it marks this part as optional (linuxrc will not abort the installation if it is missing).

## Usage examples

- load `foo` and `bar` extensions

```
extend=foo,bar
```

- load `foo` and (optionally, if it exists) `bar`
```
extend=foo,?bar
```

Note that several `extend` options override each other. So don't do `extend=foo extend=bar`, rather do `extend=foo extend=+bar` if really necessary.

### DIFF
--- a/file.c
+++ b/file.c
@@ -323,6 +323,7 @@ static struct {
   { key_zram,           "zram",           kf_cmd_early                   },
   { key_zram_root,      "zram_root",      kf_cmd_early                   },
   { key_zram_swap,      "zram_swap",      kf_cmd_early                   },
+  { key_extend,         "Extend",         kf_cfg + kf_cmd                },
 };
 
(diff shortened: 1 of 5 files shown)


lslezak (commented):
I'm not an linuxrc expert but this looks reasonable to me...

shundhammer (approved):
LGTM

PR #251, opened 2021-03-17
fix rescue image in usrmerge case (boo#1029961)

lnussel:
Do not create directories in / that are subject to usrmerge

### DIFF
--- a/linuxrc.c
+++ b/linuxrc.c
@@ -303,8 +303,8 @@ void lxrc_change_root()
   slist_t *sl;
   char *argv[3] = { };
   char *dirs[] = {
-    "bin", "boot", "etc", "home", "lib", "run",
-    "media", "mounts", "mounts/initrd", "mnt", "parts", "parts/mp_0000", "proc", "sbin",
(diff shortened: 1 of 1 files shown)


lnussel:
could be that the code is not needed at all. after all it doesn't create lib64 either

wfeldt:
Shouldn't that be a run-time decision, not compile-time?

lnussel:
depends on the use case. Do you expect linuxrc to load legacy images?

wfeldt:
Yes, linuxrc typically covers a few releases as there are strange backport requests and also people merging products (think of microos mixing sle15-sp2 and sp3).

And also I'd like to develop and compile on a different system than the test system. :-)

lnussel:
ok in that case I'd probably not create those dirs there at all. They will be in one of the images anyway

wfeldt:
Possibly; it's very old code.

lnussel:
confirmed rescue image still works without those dirs created by linuxrc

The travis fail looks unrelated

wfeldt (approved):
let's try it

PR #253, opened 2021-04-09
fix handling of modules in driver updates (bsc#1184550)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1184550
- https://trello.com/c/iroG0VHx

Kernel modules in driver updates are not handled correctly. Due to the `*.ko` to `*.ko.xz` extension change.

## Solution

Be more careful and anticipate either variant in driver updates.


shundhammer (approved):
LGTM

PR #257, opened 2021-04-21
rescue: leave linking modules to prep script

lnussel:
modules and firmware need to belinked from either /lib or /usr/lib
depending on usrmerge. Leave that to the script in installation-images


lnussel:
needs https://github.com/openSUSE/installation-images/pull/498

wfeldt (approved):

PR #259, opened 2021-04-26
fix handling of modules.order in driver updates (bsc#1184550)

wfeldt:
## Task

Port https://github.com/openSUSE/linuxrc/pull/258 to master branch.

## Original problem

- https://bugzilla.suse.com/show_bug.cgi?id=1184550
- https://trello.com/c/J127YQCF

Driver updates may contain a `modules.order` file listing the to-be-loaded modules, one per line, without '.ko' or '.ko.xz' extension.

First, all modules are unloaded in reverse order with `rmmod`, then they are updated, and then re-loaded with `modprobe`.

The modules are not required to be part of the driver update - in this case the entry just triggeres a module load.

The described process was not implemented correctly. In some cases symlinks to non-existing modules were created - which
essentially unloaded and then removed the module from the installation system.

## Solution

Fix the implementation.

Also, change `modules.order` handling to allow entries in any form (with and without explicit module extension).

### DIFF
--- a/util.c
+++ b/util.c
@@ -846,7 +846,9 @@ void add_driver_update(char *dir, char *loc)
       if(!util_check_exist(buf2) && (f = fopen(buf2, "w"))) {
         while((de = readdir(d))) {
           if(mod_basename_len(de->d_name)) {
-            fprintf(f, "%s\n", de->d_name);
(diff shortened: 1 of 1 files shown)


shundhammer (approved):
LGTM

PR #264, opened 2021-06-09
prevent memory corruption while iterating through network device list (bsc#1187014)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1187014
- https://trello.com/c/IbMHDdRx

If a network URL passed to linuxrc is not valid, linuxrc will iterate trough all available network interfaces trying them in turn to read the data. During this iteration, a memory corruption may happen.

## Background

linuxrc holds all hardware probing data in `config.hd_data` and the iteration goes over a linked list anchored there.

During the iteration linuxrc may implicitly trigger an `update_device_list` function which will update the hardware data in `config.hd_data` - but only if udev events have been processed since the last call to `update_device_list`.

Unfortunately, udev events are not only generated by new hardware during this period - and if this happens, it completely breaks the iteration loop.

## Solution

Introduce a locking mechanism blocking updates of `config.hd_data` if needed (i.e. during the loop).

### DIFF
--- a/global.h
+++ b/global.h
@@ -460,6 +460,7 @@ typedef struct {
   unsigned autoyast_passurl:1;	/**< pass autoyast url unmodified on to yast */
   unsigned device_auto_config:2;	/**< run s390 device auto-config (cf. bsc#1168036) */
   unsigned device_auto_config_done:1;	/**< set after s390 device auto-config has been run */
+  unsigned lock_device_list;	/**< prevent device list updates if != 0 */
   struct {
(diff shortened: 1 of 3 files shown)


jreidinger (approved):

jreidinger:
looks good code-wise. Travis looks kind of dead. maybe priviledged run is needed? Or maybe also convert to github actions like rest?

wfeldt:
github actions, but not really urgent

PR #272, opened 2021-07-13
fix handling of query parameters (bsc#1169492, bsc#1188186)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1188186

linuxrc may still loose query paraters in URLs. This can happen if the URL is processed **internally** and the URL is re-calculated from its parsed parts.

The older fix referenced below handles only the situation when the URL is passed on to external tools.

## See also

First attempt:

- https://bugzilla.suse.com/show_bug.cgi?id=1169492
- https://github.com/openSUSE/linuxrc/pull/222

### DIFF
--- a/url.c
+++ b/url.c
@@ -894,7 +894,7 @@ void url_add_query_string(char **buf, int n, url_t *url)
  *
  * format:
  *   0: for logging
- *   1: without query part
+ *   1: with non-standard query part
  *   2: with device
  *   3: like 2, but remove 'rel:' scheme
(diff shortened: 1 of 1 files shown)


dgdavid (approved):

anthr76:
Any chance on having this merged?

wfeldt:
I was waiting for feedback in the referenced bug.

But let's merge anyway to move this forward.

PR #273, opened 2021-08-30
Add support for gz and zst compressed modules

mwilck:
Add .ko.gz and .ko.zst to the mod_extensions array. Thanks for
having coded this with future extension in mind!

### DIFF
--- a/module.c
+++ b/module.c
@@ -46,6 +46,8 @@
 // list ends with an empty entry
 const mod_extensions_t mod_extensions[] = {
   MOD_EXT(".ko.xz"),
+  MOD_EXT(".ko.gz"),
+  MOD_EXT(".ko.zst"),
   MOD_EXT(".ko"),
   { }
 };


wfeldt:
Thanks!

Is this just prophylactic or do kernel packages using zstd already exist / are planned for?

mwilck:
At the moment, it's prophylactic. Prophylactic pactches are also on the way for suse-module-tools and the kernel itself. I have started the discussion on switching the kernel package to zstd on the suse kernel ML recently. Adding support to linuxrc would be one prerequisiste.

wfeldt:
mkdud and mksuse will also need adjustments.

Please keep in mind that whenever this is going to go into SLE, a bug or Jira number will be needed.

wfeldt (approved):
lgtm

mwilck:
So far there's just jsc#PM-2971.

PR #276, opened 2021-10-04
default to using zram (bsc#1191203, #bsc1189707)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1189707
- https://bugzilla.suse.com/show_bug.cgi?id=1191203

Use zram during installation.

This was already implemented for SP3 but not made the default. To avoid further memory shortage related issues, make zram usage the default in SP4.


joseivanlopez (approved):
LGTM

PR #279, opened 2021-11-25
allow option matching based on patterns (jsc#SLE-21308)

wfeldt:
## Task

Port https://github.com/openSUSE/linuxrc/pull/278 to SLE15-SP4.

## Original task

- https://trello.com/c/RfFcN9P7
- https://jira.suse.com/browse/SLE-21308

YaST can be controlled via a number of different environment variables (boot options) during installation. They would show up in the bootloader configuration of the target system - which is usually not wanted.

## Solution

This can be prevented by making these options 'known' to linuxrc. This has the added benefit that these settings show up in `/etc/install.inf` and they can also be set not only as boot options but also via configuration files (e.g. driver updates, autoyast config).

For this, the [`ptoptions`](https://en.opensuse.org/SDB:Linuxrc#p_ptoptions) config option has been extended to allow pattern matching for environment variable names.

For example `ptoptions=FOO_*` would make linuxrc recognize all options starting with the string `FOO_`.

## See also

- related installation-images change: https://github.com/openSUSE/installation-images/pull/547

### DIFF
--- a/file.c
+++ b/file.c
@@ -6,13 +6,14 @@
  *
  */
 
-#define _GNU_SOURCE		/* getline */
+#define _GNU_SOURCE		/* getline, fnmatch */
 
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
 #include <unistd.h>
+#include <fnmatch.h>
 #include <sys/mount.h>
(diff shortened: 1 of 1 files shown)


ancorgs (approved):
LGTM. Especially since this was already approved at #278

PR #280, opened 2022-01-12
enable display dialog for all architectures

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1193910
- https://jira.suse.com/browse/SLE-18632
- https://trello.com/c/LX0JUO1E

(1) This dialog:

```
Select the display type.

0) <-- Back <--
1) X11               
2) VNC               
3) SSH               
4) ASCII Console          

>  4
```

is misleading and should be reworked (you get the graphical Qt UI choosing 4).

(2) The dialog should be shown on `/dev/tty1` if a framebuffer exists on s390x (and not on the serial line).

## Solution

1. Enable the dialog on all architectures. For better testing and it might be useful anyway. You get the dialog by adding `netsetup=display` to the boot options (on s390x this is the default setting).
2. Switch to tty1 if a framebuffer device is available after udevd has loaded modules.
3. Adjust menu:
    - rearranged, more common choices are at the top
    - added new GUI entry
    - renamed ASCII console -> Text-based UI, X11 -> remote X11
    - fixed workflow so that you can actually abort entering passwords and get back to this menu
4. To be on the safe side, there is a new option `switchtofb` that can be used to disable step 2. in case it causes trouble.

```
Select the display type.

0) <-- Back <--
1) Graphical UI          
2) Text-based UI          
3) VNC               
4) SSH               
5) Remote X11           

> 1
```

### DIFF
--- a/dialog.c
+++ b/dialog.c
@@ -97,9 +97,10 @@ struct {
   { di_extras_quit,      "Quit linuxrc"         },
   
   { di_display_vnc,    	 "VNC"	          },
-  { di_display_x11,  	 "X11"	          },
+  { di_display_x11,  	 "Remote X11"	          },
   { di_display_ssh,      "SSH"	          },
-  { di_display_console,  "ASCII Console"	  },
+  { di_display_console,  "Text-based UI"	  },
+  { di_display_qt,       "Graphical UI"	  },
   
   { di_390net_osa,	 "OSA-2 or OSA Express"           },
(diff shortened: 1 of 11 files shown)


jreidinger (approved):

wfeldt:
replaced by https://github.com/openSUSE/linuxrc/pull/282

PR #286, opened 2022-02-25
do not leave repository mounted when starting yast (bsc#1196061)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1196061
- https://trello.com/c/bC5yVZas

linuxrc traditionally leaves the installation medium mounted when starting yast. This is not strictly necessary.

Do not do this, as this interferes with YaST's media handling and extension loading via `extend`.


shundhammer (approved):
LGTM

PR #297, opened 2022-05-24
add EDID boot option to specify arbitrary display sizes (bsc#1199020)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1199020

When dealing with HiDPI setups it's hard to reproduce issues as you normally don't have exactly the same monitor as the reporter.

## Solution

This patch adds the `edid` boot option allowing you to define the display size. This works by constructing an EDID firmware record that is subsequently read by the `drm` kernel module.

- https://en.opensuse.org/SDB:Linuxrc#p_edid

## Examples

```
# pretend you have a 4k display (default to 96 dpi)
edid=3840x2160

# pretend you have a 4k display with 150 dpi
edid=3840x2160,150

# pretend you have a 4k display with a size of 697 mm x 392 mm
edid=3840x2160,697x392
```

## Notes

1. This is mainly intended as debug option to be used in virtual environments. Your real monitor may or may not like the EDID data created here.
2. The display width and height **must be below 4096** - so no 8k display specs possible.
3. Obviously, this does not work if you don't use the kernel drm interface to access your video hardware.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ SRC	= $(filter-out inflate.c,$(sort $(wildcard *.c)))
 INC	= $(wildcard *.h)
 OBJ	= $(SRC:.c=.o)
 
-SUBDIRS	= mkpsfu
+SUBDIRS	= mkpsfu edid-write
 
 .EXPORT_ALL_VARIABLES:
 .PHONY:	all clean install libs archive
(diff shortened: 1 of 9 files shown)


joseivanlopez (approved):
LGTM

Great feature!

Issue #296, opened 2022-05-23
SLP:/ URLs doesn't works with DUD images.


PerryWerneck:
Unable use slp:/ URLs for DUD images.

PR #298, opened 2022-06-03
Add support to scheme SLP: when getting DUD files.

PerryWerneck:
enable use of slp protocol to detect servers for the DUD images.

Related with issue #296


PerryWerneck:
The original problem was solved moving the 'dud' image to the boot media.

PR #310, opened 2023-01-17
support more general wicked firmware devices interface (jsc#PED-3118, jsc#PED-967)

wfeldt:
## Task

- https://jira.suse.com/browse/PED-967
- https://jira.suse.com/browse/PED-3118

wicked now has a more general extension that lists firmware handled network interfaces (not just ibft).

From linuxrc's perspective all these interfaces should just be ignored (not offered for configuration). They are set up by a `wicked ifup all` call.

## See also

- https://github.com/openSUSE/installation-images/pull/620

### DIFF
--- a/file.c
+++ b/file.c
@@ -315,7 +315,8 @@ static struct {
   { key_sethostname,    "SetHostname",    kf_cfg + kf_cmd_early          },
   { key_debugshell,     "DebugShell",     kf_cfg + kf_cmd + kf_cmd_early },
   { key_self_update,    "SelfUpdate",     kf_cfg + kf_cmd                },
-  { key_ibft_devices,   "IBFTDevices",    kf_cfg + kf_cmd                },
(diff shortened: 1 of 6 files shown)


ancorgs:
Not sure if `WithNVME` is a good name. I guess it was chosen by similarity with `WithISCSI`. But if I'm not mistaken, this case is actually quite different.

`WithISCSI` implies iSCSI configuration is desired. That triggers a manual process in YaST to setup one (or many) iSCSI targets. So it implies iSCSI is used in general, no matter whether there is an iBFT (iSCSI Boot Firmware Table) with entries or not.

The new parameter is about NBFT (that is booting from the information available at the NVMe Boot Firmware Table) and it set if there are entries at the NBFT. Something that should not trigger or need any manual setup. So it's not a 1:1 to `WithISCSI` and it would be better to use "NBFT" as part of the name (instead of the more generic "NVME"). Isn't it?

Issue #322, opened 2023-07-10
Malformed packet when using SLP with customized service name.


PerryWerneck:
When we change the SLP service name using URL SLP:/?service=whathever the package sent is malformed due to invalid value on the field 'Service type length'  (fixed as 0014). A workaround can be done keeping the name of the custom service with the same length.

`
	0x0030:  0000 0000 0006 0002 656e 0000 [0014] 7365  ........en....se
	0x0040:  7276 6963 653a 696e 7374 616c 6c2e 7370  rvice:install.sp
	0x0050:  3400 0764 6566 6175 6c74 0058 2826 287c  4..default.X(&(|
		
	0x0030:  0000 0000 0009 0002 656e 0000 [0014] 7365  ........en....se
	0x0040:  7276 6963 653a 696e 7374 616c 6c2e 7375  rvice:install.su
	0x0050:  7365 0007 6465 6661 756c 7400 5828 2628  se..default.X(&(

`

PR #323, opened 2023-07-10
Fixing Length of service type list field on Service Request package.

PerryWerneck:
Fixes the length of the service type field on Service Request package, it was fixed on code and causing 'malformed packet' when using a customized service name (Issue #322)


wfeldt:
Replaced by https://github.com/openSUSE/linuxrc/pull/324.

